PR #46, opened 2019-08-11
Group by version for libs

takahirom:
I want to group libs like this.
image

If you like this, please merge.

### DIFF
--- a/plugin/src/main/kotlin/de/fayard/BuildSrcVersionsTask.kt
+++ b/plugin/src/main/kotlin/de/fayard/BuildSrcVersionsTask.kt
@@ -52,7 +52,7 @@ open class BuildSrcVersionsTask : DefaultTask() {
 
         val dependencyGraph = readGraphFromJsonFile(jsonInput)
 
-        val useFdqnByDefault = useFdqnFor.get().map { it.lowerSnakeCase() }
+        val useFdqnByDefault = useFdqnFor.get().map(::escapeName)
 
         val dependencies: List<Dependency> = parseGraph(dependencyGraph, useFdqnByDefault + MEANING_LESS_NAMES)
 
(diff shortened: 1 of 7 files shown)


takahirom:
Sorry I found a bug. Please let me close 

If libs inside an object. Gradle can not find the object.

```
buildscript {
    dependencies {
        classpath Libs.org_jetbrains_kotlin.kotlin_gradle_plugin
        classpath Libs.org_jetbrains_kotlin.kotlin_serialization
```

```
object Libs {
    object org_jetbrains_kotlin {
        /**
         * https://kotlinlang.org/ */
        const val kotlin_android_extensions_runtime: String =
                "org.jetbrains.kotlin:kotlin-android-extensions-runtime:" +
                Versions.org_jetbrains_kotlin
...
```

```
> No such property: org_jetbrains_kotlin for class: Libs
```

romainpiel:
That's a shame, I think this was a great feature idea. Any plan to build this @takahirom or @jmfayard ? Otherwise I'm happy to investigate into it.

jmfayard:
@romainpiel 
I would be happy to help you if you want to do that. The first step would be to investigate is what you have to change to this Kotlin code

```
// buildSrc/src/main/A.kt
object A {
   object B {
     const val hello = "Hello World!"
  }
}
``` 

so that this code works in Groovy exactly like it works in Kotlin

```kotlin
// build.gradle

// Run with ./gradlew hello
tasks.create("hello") {
  doLast {
    System.out.println("A.B.hello=" + A.B.hello)
  }
}
```

takahirom:
I think we can fix by adding package like "dependencies" for Libs object. And we should add import like "import dependenciet.Libs" for build.gradle.
But it will be breaking change.Is it ok?

jmfayard:
I doesn't have to be a breaking change because we can add a configuration option see #53 
I'm not really convinced that having packages is what we want.

jmfayard:
Blocked by https://github.com/jmfayard/buildSrcVersions/issues/90

jmfayard:
I have found out how to do it properly cc @takahirom @romainpiel 
If you are interested, see https://github.com/jmfayard/buildSrcVersions/issues/90

romainpiel:
Yay nice! Thanks @jmfayard

PR #117, opened 2019-10-25
starting point for @LouisCAD

jmfayard:
Usage: 

```bash
$ cd sample-kotlin 
$./gradlew refreshVersionsUpdates
$ cat build/dependencyUpdates/refreshVersions.txt  
```

- https://github.com/LouisCAD/Splitties/commit/9f4a85e7ecc612d290fd6e4615b3472c05aece51
- https://github.com/jmfayard/buildSrcVersions/issues/102
- https://github.com/ben-manes/gradle-versions-plugin#report-format
- also think about https://github.com/jmfayard/buildSrcVersions/issues/116


ben-manes:
Curious, what is the desired difference using LouisCAD's? It seems to taken the same approach as the versions plugin, except it hasn't suffered through the evolution of Gradle and adding of requested features. The [earliest version](https://discuss.gradle.org/t/add-report-to-display-dependency-updates/540) was merely a script task published on the Gradle forum. The core logic is tiny, so you could extract and embed directly.

The drawback of delegating to Gradle has been performance, because their team doesn't want to parallelizing metadata fetching like Maven does. That was decided years ago, briefly explored when jar downloads were parallelized, so it may be worth revisiting.

jmfayard:
Hello @ben-manes nice to have you here!
@LouisCAD initial motivation is that he is doing Kotlin Multiplatform stuff, and as you know we have this un-resolved issue with it https://github.com/ben-manes/gradle-versions-plugin/issues/334

I have not checked its solution in detail to be honest, but the core of it is at https://github.com/LouisCAD/Splitties/commit/9f4a85e7ecc612d290fd6e4615b3472c05aece51

Instead of copying the configuration "implementation" to "implementation-copy" with multiple dependencies, it creates for each dependency a new configuration for it alone, resolve it - apparently it then works - and immediatly destroy it again.

Also Louis has an incredibly complex Gradle project with Android build variants and Kotlin Multiplatform and 40 modules. With an impressive total of **13.288 Gardle taks**

https://scans.gradle.com/s/nqyakfjv6yqcu/performance/configuration

So performance is a big topic for him.

One idea we have is to let the users configure which versions should be aligned.
For example the 40 dependencies `com.louiscad.splitties:splitties.*` should all share the same version, so it's wasteful to resolve it 40 times - and maybe in multiple modules? Not sure how it works. 

Also to be honest, another problem is that we both are not that good with Groovy and prefer to write in Kotlin.

jmfayard:
replaced by https://github.com/jmfayard/buildSrcVersions/pull/124

PR #124, opened 2019-11-05
refreshVersions.useExperimentalUpdater (LouisCAD)

jmfayard:


plugin/src/main/kotlin/de/fayard/versions/extensions/GradleCompat.kt
```diff
@@ -0,0 +1,16 @@
+package de.fayard.versions.extensions
+
+import org.gradle.api.Action
+import org.gradle.api.Task
+import org.gradle.api.tasks.TaskContainer
+import org.gradle.util.GradleVersion
+
+internal inline fun <reified T : Task> TaskContainer.registerOrCreate(name: String) {
```
jmfayard: Nice! I wanted to create something like this. 
I have suggested it here https://youtrack.jetbrains.com/issue/KT-25428

PR #131, opened 2019-11-08
Version 0.8.4: use experimentalUpdater for buildSrc consumers

LouisCAD:


LouisCAD:
Misses supports for alias to plugin versions. Also, I found a way to test locally with buildSrc, so I'll make a new PR later, possibly with other improvements.

PR #149, opened 2019-11-29
refreshVersions 0.8.6

### LINKED ISSUE #144: Some plugins versions are removed by refreshVersions
Permalink to reproducing project: https://github.com/LouisCAD/update-legacy-android-project/tree/66f7e854f33593c2a990f40f8f12abbaf66f15f1

`plugin.android=3.5.0` and `plugin.de.fayard.dependencies=0.5.0` are ditched by the `refreshVersion` Gradle task while they are used.

I need to find where I can get the fact that they are used and that they get their version from `versions.properties` to include them when running `refreshVersions`.

### LINKED ISSUE #148: Ensure libraries (and Gradle plugins) publish their used version and not the placeholder
Solutions to explore:
1. `mavenPublications.all { versionMapping { allVariants { fromResolutionResult() } } }`
2. Replacing dependencies in `afterEvaluate` (or manually by function call?) with hardcoded versions
3. Using [dependency substitution](https://docs.gradle.org/current/dsl/org.gradle.api.artifacts.DependencySubstitutions.html)?

This was brought to my attention by @melix after he told me resolved version is not published by default, only the declared one.

LouisCAD:
**Known issue:** running the clean task just before in the same command will
make refreshVersions run improperly.

Content :
- Replace resolutionStrategy (for non plugins) by dependencies replacement to not publish the version placeholders. (closes #148)
- Ensure we keep track of the used plugins where the used version has been defined using the version property (closes #144)


refreshVersions/src/main/kotlin/de/fayard/versions/internal/UsedDependenciesWriting.kt
```diff
@@ -3,13 +3,22 @@ package de.fayard.versions.internal
 import de.fayard.versions.extensions.isRootProject
 import org.gradle.api.Project
 import org.gradle.api.artifacts.Dependency
+import org.gradle.api.initialization.Settings
 import org.gradle.api.internal.artifacts.dependencies.AbstractDependency
+import java.lang.Appendable
 
-internal fun Project.readDependenciesUsedInBuildSrc(): Sequence<Dependency> {
+internal fun Project.readExtraUsedDependencies(): Sequence<Dependency> {
```
jmfayard: the naming is not really clear..

refreshVersions/src/main/kotlin/de/fayard/versions/internal/UsedDependenciesWriting.kt
```diff
@@ -30,16 +39,41 @@ internal fun Project.writeUsedDependencies() {
...
+            writer.appendDependency(
+                group = dependency.group,
+                name = dependency.name,
+                version = dependency.version
+            )
         }
     }
 }
 
+internal fun Settings.clearUsedPlugins() {
```
jmfayard: naming also not really clear here

refreshVersions/src/main/kotlin/de/fayard/versions/internal/UsedDependenciesWriting.kt
```diff
@@ -30,16 +39,41 @@ internal fun Project.writeUsedDependencies() {
 
     file.bufferedWriter().use { writer ->
         allDependencies.forEach { dependency ->
-            writer.append(dependency.group)
-            writer.append(':')
-            writer.append(dependency.name)
-            writer.append(':')
-            writer.append(dependency.version)
-            writer.appendln()
+            writer.appendDependency(
```
jmfayard: One thing I want to change going forward is to do more in unit tests. 
Right now if I make changes in the plugin, I have no idea what will break in the projects you have at work. 
Not nice :)

So let have more unit tests, but how?

I'm a big believer in the "Functional core - Imperative shell" principle described here 

https://www.destroyallsoftware.com/screencasts/catalog/functional-core-imperative-shell

As it is, the function `writeUsedDependencies()` and `writer.appendDependency()` cannot be unit-tested easily. 

I would suggest to have a core function that takes a list of dependencies and return a string that corresponds to the file content (functional core) and unit test the hell out of it. 

Then the imperative shell will be just `file.writeText(dependencies.format())`

jmfayard:
Thanks @LouisCAD
We will have to talk about unit testing going forward

refreshVersions/src/main/kotlin/de/fayard/versions/internal/UsedDependenciesWriting.kt
LouisCAD: Renamed to `readPluginsAndBuildSrcDependencies`.

refreshVersions/src/main/kotlin/de/fayard/versions/internal/UsedDependenciesWriting.kt
LouisCAD: Renamed to `clearUsedPluginsList`. If you have a better idea, please let me know.

refreshVersions/src/main/kotlin/de/fayard/versions/internal/UsedDependenciesWriting.kt
LouisCAD: As to the `appendDependency` function, it's an extension on `Appendable` so you can write a file-less unit test for it using a `StringBuilder` (like in `buildString { ... }`).

That said, I agree that testability can be improved, and that functional core, imperative shell is a goo approach.

Issue #155, opened 2020-01-26
Bootstraping refreshVersions


jmfayard:
**Context**

We have a chicken-and-egg problem on how to tell Gradle that plugins versions should be found in `versions.properties`. 

**Describe the solution you'd like**

The setup for Groovy should now be:

```groovy
// settings.gradle
import de.fayard.versions.RefreshVersionsSetup
buildscript {
        dependencies.classpath("de.fayard:refreshVersions:VERSION")
}
RefreshVersionsSetup.bootstrap(settings)
```

The setup for Kotlin should now be:

```kotlin
// settings.gradle.kts
import de.fayard.versions.bootstrapRefreshVersions

buildscript {
    dependencies.classpath("de.fayard:refreshVersions:VERSION")
}

settings.bootstrapRefreshVersions()
```

**Describe alternatives you've considered**

- users write tons of boilerplate in `settings.gradle`
- the plugin copy/paste that boilerplate in the `gradle` folder
- we try to use the `plugins` block in settings.gradle

none of this is nice or works with Gradle 5-6

The best solution so far was found by @LouisCAD: using the classpath inside `settings.gradle`

https://github.com/jmfayard/refreshVersions/blob/a20c43b071a0e17eb92545710ec89df8c562796a/sample-groovy/settings.gradle#L1-L19

Biggest drawback of this solution: the bootstrap function is tied to the version of the plugin. This makes updating `refreshVersions` itself potentially unsafe, and at least not as easy as the other plugins.

LouisCAD:
Resolved in version 0.9.1

PR #159, opened 2020-02-11
Simplify setup

jmfayard:
Changelog

 🎉 Make the new updater the default
✨ Bootstraping refreshVersions #155 

The setup for Groovy should now be:

```groovy
// settings.gradle
import de.fayard.versions.RefreshVersionsSetup
buildscript {
        dependencies.classpath("de.fayard:refreshVersions:VERSION")
}
RefreshVersionsSetup.bootstrap(settings)
```

The setup for Kotlin should now be:

```kotlin
// settings.gradle.kts
import de.fayard.versions.bootstrapRefreshVersions

buildscript {
    dependencies.classpath("de.fayard:refreshVersions:VERSION")
}

settings.bootstrapRefreshVersions()
```


refreshVersions/src/main/kotlin/org/gradle/kotlin/dsl/bootstrap.kt
```diff
@@ -0,0 +1,20 @@
+package org.gradle.kotlin.dsl
```
jmfayard: that was a good tip from Nikky: no import is necessary when you put code in this package

LouisCAD: Did you try putting it in the root package?

jmfayard: @LouisCAD not necessary, things under `org.gradle.kotlin.dsl` are auto-imported in a `build.gradle.kts` file

Actually we might want to put the dependencies constants under the same package inside the `dependencies` plugin

LouisCAD: I really don't like that hack as it can be confusing regarding the source. So long as there's no way to add to the implicit imports list, I'll pick root package anytime as it doesn't give false information about the source of the symbol, and encourages investigators to search further instead of being misled.

jmfayard:
@LouisCAD regarding to the package issue
Answer from Louis Jacomet on the Gradle slack channel

> Yes it works, but no you should really not do it. You are effectively leveraging split packages.
> You end up squatting the Kotlin DSL namespace in a plugin, potentially causing risks of collision with others attempting the same or even Gradle adding new things.

I think this equally applies to using the root package, so I moved it simply to `de.fayard.versions`

LouisCAD:
I think if we have a name that is very unique for the bootstrap symbols, using the root package is less of an issue as there would be no clash with future Gradle APIs, buildSrc code or other dependencies using root package as well.

jmfayard:
We don't need to use the root package, IntelliJ auto-improts flawlessly the extension fuction if you don't put the import.

jmfayard:
@LouisCAD please review the last commit who generate properties for all deps
https://github.com/jmfayard/refreshVersions/pull/159/commits/2173c74c7893847dcfda31e4b559f679486fc887

LouisCAD:
I merged master into this branch now that there's a file format and exposed API for version keys (#161), and I fixed the conflicts.

I'm in the process of reviewing locally, fixing what needs to be fixed, before this can be merged and released.

LouisCAD:
I noticed `versions.properties` is now filled with all versions, but for the ones that are hardcoded, the latest version is put, not the version that is actually used with the other candidates. That happens because the plugin wasn't designed for the change of getting all the versions.

I'll work on a solution for that in this PR, or revert that part, so it's being worked on in another PR, after this one is merged.

LouisCAD:
Also, the changes broke dependencies rewrite as you can see the version placeholder stays in the publication as the GitHub actions failure shows.

refreshVersions/src/main/kotlin/de/fayard/versions/internal/VersionsPlaceholdersReplacement.kt
```diff
@@ -30,12 +32,23 @@ internal fun Project.setupVersionPlaceholdersResolving() {
...
+                val dependenciesToReplace = dependencies.filter { it.version == versionPlaceholder }
                 removeAll(dependenciesToReplace)
-                for (dependency in dependenciesToReplace) {
+
+                if (dependencies.any { it.version != versionPlaceholder }) {
+                    val warnFor = (dependencies - dependenciesToReplace).take(3).map { it.name }
+                    logger.warn(""":${project.name}:${configuration.name} found hardcoded dependencies $warnFor   See https://github.com/jmfayard/refreshVersions/issues/160 """)
+                }
+
+                for (dependency in dependencies) {
```
LouisCAD: This currently messes up the versions.properties file and brings two sources of truth because the currently used version is not retrieved.

Let's change that in another PR, all while adding unit tests along the way.

PR #162, opened 2020-02-18
Dependencies plugin bootstrap

LouisCAD:
Configuring the dependencies plugin is now explicit, automatically applies refreshVersions.

An inconsistency bug where the rules would not be used has also been fixed.

Finally, this bring an API to add more rules by giving the content of the version key rules files, as done for the dependencies plugin.


jmfayard:
LGTM @LouisCAD

LouisCAD:
I'll fix the groovy sample in minutes and merge it

PR #165, opened 2020-02-27
Add migration facilities

LouisCAD:
This PR adds a new `bootstrapRefreshVersionsAndDependencies` function that does exactly what its name describes.

This brings a new interactive `migrateToRefreshVersionsDependenciesConstants` Gradle task in the root project. It's advised to run it with the option `--console=plain` to get rid of the irrelevant execution progress bar.

The task is not perfect, but is enough to make migration easy and safe enough.

This PR also fixes two bugs that have been seen in #164


dependencies/src/main/kotlin/de/fayard/dependencies/internal/MigrationToDependenciesConstants.kt
```diff
@@ -0,0 +1,98 @@
...
+import org.gradle.api.Project
+import org.gradle.api.artifacts.Configuration
+import org.gradle.api.artifacts.ExternalDependency
+import kotlin.coroutines.coroutineContext
+
+internal fun Configuration.shouldBeIgnored(): Boolean {
+    return name.startsWith(prefix = "_internal") // Real-life example: _internal_aapt2_binary (introduced by AGP)
+        || name in ignoredConfigurationNames || name.startsWith('-')
+    //TODO: If unwanted configurations still get through, we can filter to known ones here, like
+    // implementation, api, compileOnly, runtimeOnly, kapt, plus test, MPP and MPP test variants.
```
jmfayard: I would rather have a whitelist approach.

dependencies/src/main/kotlin/de/fayard/dependencies/internal/MigrationToDependenciesConstants.kt
```diff
@@ -0,0 +1,98 @@
...
+import kotlinx.coroutines.*
+import org.gradle.api.Project
+import org.gradle.api.artifacts.Configuration
+import org.gradle.api.artifacts.ExternalDependency
+import kotlin.coroutines.coroutineContext
+
+internal fun Configuration.shouldBeIgnored(): Boolean {
+    return name.startsWith(prefix = "_internal") // Real-life example: _internal_aapt2_binary (introduced by AGP)
+        || name in ignoredConfigurationNames || name.startsWith('-')
+    //TODO: If unwanted configurations still get through, we can filter to known ones here, like
```
jmfayard: What about all the cases that we don't know about?
For example, the gradle-versions-plugin add tons of configurations ending with `-copy`
I would rather assume that we don't know all the possible ways that everyone is using Gradle and have a whitelist approach

refreshVersions/src/main/kotlin/de/fayard/versions/internal/VersionsPlaceholdersReplacement.kt
```diff
@@ -146,9 +106,26 @@ internal fun String.isAVersionAlias(): Boolean = startsWith("version.") || start
 
 private val lock = Any()
 
-private fun Project.versionsPropertiesFile(): File {
+internal fun Project.versionsPropertiesFile(): File {
```
jmfayard: Can you rename to `readOrCreateVersionsPropertiesFile()`

jmfayard:
@LouisCAD 
Naming isssue: wouldn't it be better if all or your tasks start with the same prefix from a usability persperctive?

- `$ ./gradle refreshVersions`
- `$ ./gradle refreshVersionsBootstrap`
- `$ ./gradle refreshVersionsMigrateDependenciesConstants`

LouisCAD:
It wouldn't, because you then wouldn't be able to type ./gradlew rV to invoke the refreshVersions task which is the only one I expect users to run on a daily basis from their work machine.

dependencies/src/main/kotlin/de/fayard/dependencies/internal/MigrationToDependenciesConstants.kt
LouisCAD: Both blacklists and whitelists are not viable because you'll keep on getting new unexpected configuration names. For example, KMPP allows dynamic naming of configurations.

Regarding the `gradle-versions-plugin`, do you expect people to keep it as they add refreshVersions to their project? Maybe we could advise to replace it, or even detect its presence?

jmfayard:
@LouisCAD I investigated a bit and I do prefer a common prefix - for the exact naming of the task, I let you choose whatever makes sense.

I think zsh has auto-completion for Gradle tasks - although I don't use it, but that's not the only use case.

By having a common prefix, tasks are grouped together just by the virtue of being sorted alphabetically. 

That works both in the therminal and in the IDE:

check-build_yml_—_refreshVersions

And `Run Gradle task` does provide auto-completion based on a common prefix

refreshVersions____IdeaProjects_refreshVersions_refreshVersions__-_____src_main_kotlin_de_fayard_RefreshVersionsPlugin_kt__refreshVersions_main__and_main_yml_—_android-kiss

LouisCAD:
Last comment resolved over the phone. Merging for release, further improvements will come later.

PR #180, opened 2020-04-09
New naming & publishing with dev versions on bintray

LouisCAD:
"refreshVersions" is now called "refreshVersions-core",
and "dependencies" is now called "refreshVersions"

There's also a new `Releasing.main.kts` script for safe semi-automated releases, and we can also push dev releases by just changing the version now, handy for quick-checks.


LouisCAD:
The plan is to add an interactive script/task for the version update and publishing process in this PR before merging it.

PR #187, opened 2020-05-12
Add grouping rule for the Alchemist simulator

DanySK:
## Issue

The [Alchemist Simulator](https://alchemistsimulator.github.io/) is usually used via Gradle and refreshVersions is in use in the reference projects.

- [ ] Refactor
- [ X ] Feature
- [ ] Bug Fix
- [ ] Documentation Update

## Description

This PR allows for using a single version identifier for all the components of the simulator, which must get pulled at once.

### DIFF
--- a/dependencies/src/main/resources/refreshVersions-rules/other-version-alias-rules.txt
+++ b/dependencies/src/main/resources/refreshVersions-rules/other-version-alias-rules.txt
@@ -6,3 +6,6 @@ com.squareup.*:*
 
 io.ktor:*
    ^^^^
+
+it.unibo.alchemist:alchemist-*
+         ^^^^^^^^^


DanySK:
@LouisCAD I hope I correctly understood from your video :)

LouisCAD:
From the video, I showed how it's done internally, and how to add it in your project.

The word `alchemist` is quite generic and can clash with other libraries (like [this one](https://github.com/shuleisanshi/alchemist/blob/b5862e6ea76be876d676f3b57a911c18775e69d5/pom.xml#L12)).

Dealing with such clashes has the potential to be a nightmare for maintaining, so I cannot merge this PR.

That doesn't prevent you from using that rule in your project with the optional parameter for the bootstrap function in `settings.gradle.kts`.

DanySK:
Uhmm... Wouldn't that be used by default with the full group name? Or am I
misunderstanding how grouping is supposed to work?

Would "it.unibo.alchemist" in place of just "alchemist" get accepted?

Anyway, please provide a general policy on grouping and clear rules for
valid proposals. Right now it seems to me that choices of what can be
reified as a group by default is entirely arbitrary.

I also maintain the Protelis programming language, is that a suitable
candidate?

LouisCAD:
I had planned to write such a policy, but that also takes time.
I think you've seen dependencies gravitate around the Kotlin and Android ecosystems, but even then, I cannot maintain a list of everything because the risk is breaking people builds on incompatible changes. That's why I stick to libraries I'm familiar with, and I'm not considering deviating from that route, but I'd encourage people to write their Gradle plugins, and library authors can as well, which means for example that Spring could make a first party one.

DanySK:
The policy is reasonable, I understand it.
Is there a guideline on the preferred way to write such plugins? I suspect that if everyone goes by its path, we might end up with cross-incompatible plugins.

My suggestion would be to suggest writing plugins exposing methods such as:
`groupXXXversions(): String` to be passed directly to the boostrap method.

Any adverse effect of this design choice?
Is there any plugin out there already?
Otherwise, I can make some, and if you agree on the structure we could use them as examples.

LouisCAD:
You mentioned an interesting point when it comes to cross-compatibility.

Currently, refreshVersions will just pick the longest rule. This might not be the best option.
Regardless, it'll be best to put the dependencies constants in a package that is unique to the plugin to avoid any major issues.

I'll keep that in mind when I'll improve design for plugin authors and provide an example.

With all of the above, I think you have everything you need to make your own plugin, though #189 might affect you.

FYI, I should resume my development efforts on this project in about less than a month (mid-June), and this will be on the list.

PR #206, opened 2020-07-18
Update testing for Kotest

sksamuel:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Documentation Update

## Description

Added the dependencies that users would want to add. I've dropped the console dependency as that's going away in the next week or two. I left in the KotlinTest block as I didn't know if you were going to remove those entirely or leave in the deprecated versions.


plugins/dependencies/src/main/kotlin/dependencies/Testing.kt
```diff
@@ -102,6 +104,62 @@ object Testing {
...
+
+        object Extensions {
+            private const val artifactPrefix = "$artifactBase-extensions"
+
+            const val spring = "$artifactPrefix-spring:_"
+            const val koin = "$artifactPrefix-koin:_"
+            const val allure = "$artifactPrefix-allure:_"
+            const val testContainers = "$artifactPrefix-testcontainers:_"
+            const val http = "$artifactPrefix-http:_"
+            const val mockserver = "$artifactPrefix-mockserver:_"
```
LouisCAD, line 145: If we have `piTest` and `testContainers`, shouldn't we have `mockServer` here instead of `mockserver`, for case consistency?

sksamuel: I suppose I thought of mockserver as a single word but please feel free to adjust anything to match the style of this project.

LouisCAD:
Thank you very much for you contribution! I noted the detail regarding the console artifact in the commit message, that was helpful.

sksamuel:
You're welcome, thank you and @jmfayard for a great project.

Issue #208, opened 2020-07-21
Add support for LeakCanary


OscarSpruit:
**Is your feature request related to a problem? Please describe.**
The plugin doesn't support LeakCanary dependencies yet.
 
**Describe the solution you'd like**
Would be nice to add support, as requested in https://github.com/square/leakcanary/issues/1895.

OscarSpruit:
Closing this issue as it has been resolved with #209

PR #209, opened 2020-07-21
Add Square.leakCanary constant

OscarSpruit:
## Issue
#208 

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Documentation Update

## Description
Added `Square.leakCanary` constant to support the three most relevant LeakCanary dependencies.


LouisCAD:
Thanks for the PR! I discovered a few more LeakCanary dependencies by reviewing its documentation, and I think these are worth adding, so I added them.

OscarSpruit:
Nice! I believe there are even more, but I'm not sure which are needed and which not 🤔

LouisCAD:
Hello @pyricau, can you tell us if other artifacts are missing?

I looked at [maven central search here](https://search.maven.org/search?q=g:com.squareup.leakcanary), but I'm unsure which ones should be filtered out because not designed for public use (but only as transitive dependencies of public-use designed artifacts).

LouisCAD:
Merged this. Will add links in KDoc, and the gradle plugin dependency in upcoming commits.

Issue #215, opened 2020-08-13
> java.net.MalformedURLException: unknown protocol: s3


saadfarooq:
I use a maven repo from Amazon S3, which `refreshVersions` apparently can't handle.

```
maven {
            name "MavenS3"
            url "s3://my-custom/maven"
            credentials(AwsCredentials) {
                accessKey "{ACCESS_KEY}"
                secretKey "{SECRET_KEY}"
            }
        }
```

Works correctly after removing that.

LouisCAD:
Can you link the doc of  `AwsCredentials`?

saadfarooq:
https://docs.gradle.org/current/userguide/declaring_repositories.html#sec:supported_transport_protocols

AND 

https://docs.gradle.org/current/dsl/org.gradle.api.credentials.AwsCredentials.html

The issue is the s3 protocol though, without credentials it would just fail I guess, but right now it doesn't know what to do with the URI.

PR #216, opened 2020-08-14
Temporary fix for issue#215, MalformedURLException when using S3 maven repo

saadfarooq:
## Issue

https://github.com/jmfayard/refreshVersions/issues/215

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Documentation Update

## Description
Plugin fails when using S3 maven repos because the repo address doesn't conform to URL specs (it's a URI). The URI class doesn't have an easy way to `readText` from any URI, so for the time being ignoring the S3 repos. A proper fix might need to involve the Gradle provided repository handler API's since S3 repos will usually be associated with credentials as well and simply opening a connection to the URI is unlikely to solve that, but I don't have enough expertise in the Gradle APIs to provide more than this temporary fix.

### DIFF
--- a/refreshVersions/src/main/kotlin/de/fayard/versions/internal/LatestVersionsRetrieval.kt
+++ b/refreshVersions/src/main/kotlin/de/fayard/versions/internal/LatestVersionsRetrieval.kt
@@ -9,6 +9,8 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.invoke
+import java.net.MalformedURLException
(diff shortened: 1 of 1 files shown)


LouisCAD:
Hi, I don't think this issue can happen anymore in the latest dev version (if you want to try it, it's the version `0.9.5-dev-012`, new docs are on the develop branch).

saadfarooq:
Doesn't seem to be fixed.

```
 Can not use getCredentials() method when not using PasswordCredentials; please use getCredentials(Class)
```

Think the plugin is assuming username/password credentials only whereas this is class based auth (AwsCredentials).

PR #229, opened 2020-08-26
✚ plural type-safe way of adding dependencies

jmfayard:
Augment the Kotlin DSL for the block `dependencies { }`
with its plural forms `implementations`, `apis` and `testImplementations`.

For example:

```
dependencies {
  implementations(KotlinX.coroutines.core, KotlinX.coroutines.android)
}
```

This is both more concise and more type-safe.
`implementation(KotlinX.coroutines)` is wrong but would fail only at runtime because the method accept an object as parameter

Voice____IdeaProjects_android_Voice__-_build_gradle_kts___data_

### DIFF
--- a/plugins/core/src/main/kotlin/org/gradle/kotlin/dsl/dependencies.kt
+++ b/plugins/core/src/main/kotlin/org/gradle/kotlin/dsl/dependencies.kt
@@ -1,0 +1,27 @@
+package org.gradle.kotlin.dsl
+
+import org.gradle.api.artifacts.dsl.DependencyHandler
+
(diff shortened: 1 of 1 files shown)


LouisCAD (changes requested):
Having multiple dependencies declarations makes eye scanning harder, as you need to juggle back and forth in addition to scanning down, and you lose the alignment.

The type safety is a nice thing though I like it. I think this is an area that deserves exploration as your proposal is not the only possible solution.

Maybe we should open an issue about that, or discuss this in another channel?

BTW, for future PRs, let's target the develop branch so people always browse the code reflecting the latest stable release by default.

Issue #230, opened 2020-08-27
Square.sqlDelight.coroutinesExtensions misses the version placeholder


LouisCAD:
It's hard to make an automated test that'd prevent future recurrence because we have dependencies that explicitly require no version because they are used with BoM (e.g. Firebase).

Maybe we can keep a list of them to see when it grows.

jmfayard:
That's done

LouisCAD:
And that's now released in 0.10.0

PR #231, opened 2020-08-28
POC: bring back the task ./gradlew buildSrcVersions #230 

jmfayard:
Add placeholder to Square.sqlDelight.coroutinesExtensions 

Add a unit test that checks that the placeholder is not missing 
unless the dependency belongs to a whitelist of dependencies using the BoM


LouisCAD:
Nice! My dream is coming true, I'll review this PR this weekend.

LouisCAD:
Would be great to bring back that in another PR:
> Add a unit test that checks that the placeholder is not missing
unless the dependency belongs to a whitelist of dependencies using the BoM

Issue #227, opened 2020-08-24
v0.9.5 no longer automatically loads the latest version, is this wanted?


DanySK:
I noticed that, in case a version is unspecified in the `version.properties` file, refreshVersions 0.9.5 does not bind it to the latest it can, but fails with a message like:
```
        * What went wrong:
        Execution failed for task ':refreshVersions'.
        > Couldn't resolve version for ModuleId(group=com.github.kittinunf.fuel, name=fuel-coroutines)
```
I used this auto-select-latest feature to manually migrate up to date stuff quickly, was it dropped purposefully?
It can make it a little harder to migrate, especially when there are unknown custom version key rules, or when they change.

DanySK:
It is strange because you can provide it with a non-existent version, and as far as the entry is correctly entered the system works as intended :open_mouth:

DanySK:
To reproduce:

`settings.gradle.kts`
```kotlin
import de.fayard.refreshVersions.bootstrapRefreshVersions

buildscript {
    repositories { gradlePluginPortal() }
    dependencies {
        classpath("de.fayard.refreshVersions:refreshVersions:0.9.5")
    }
}
bootstrapRefreshVersions()
```

`build.gradle.kts`
```kotlin
plugins {
    `java-library`
}

repositories {
    mavenCentral()
}

dependencies {
    implementation("org.scala-lang:scala-compiler:_")
}
```

This build fails, while adding

`versions.properties`
```java
version.org.scala-lang..scala-compiler=foo
```

Makes the build work. I believe it is a bug, or at least looks like that from an user perspective.

LouisCAD:
I think it's a bug/regression. I'll take a look later.

jmfayard:
Major bug that I can reproduce all the time, should be fixed with https://github.com/jmfayard/refreshVersions/pull/232

DanySK:
@jmfayard before merging maybe it'd make sense to add a regression test, this looks quite a delicate feature not to have covered

jmfayard:
I agree, best would be to add integratino tests based on **gradle testKit**
https://github.com/jmfayard/refreshVersions/issues/59

DanySK:
[...]

PR #232, opened 2020-08-28
🐛 #227 refreshVersions fails when a property is missing in versions.properties

jmfayard:
Before:
If the key for a property was missing, refreshVersions 0.9.5 failed with

`> Couldn't resolve version for ModuleId(group=XXX)`

After: the new key is generated
```
// versions.properties
version.org.scala-lang..scala-compiler=2.13.3
```


LouisCAD:
I'll check this against the history this weekend to see what changed exactly and if it's the right fix (but I think you found it right where it is).

jmfayard:
Can this be merged?

LouisCAD:
Not yet, I need to review it and test it locally. It's in my top priorities.

LouisCAD:
@jmfayard I still have a hard time reproducing the issue without that PR. Can you find instructions that reproduce it all the time so I can check the difference and ensure we're fixing it correctly?

LouisCAD:
I tested locally and I can still reproduce the issue with this PR deployed on mavenLocal.

To test it, I added the following into `sample-kotlin` dependencies:

```
implementation("org.kodein.di:kodein-di:_")
```

and on Gradle import in IDEA, I get this error message:

```
Could not resolve: org.kodein.di:kodein-di:7.1.0
```

I didn't sync again because if I did, the error would go away, and the issue would no longer be reproducible for that dependency, even after removing the entry in `versions.properties`.

As long as I keep removing the `versions.properties` entry before Gradle sync, I can reproduce the issue.

We need to investigate deeper.

LouisCAD:
Okay, I found the root cause of issue #227… and while I'm very happy that I'll be able to fix it, I'm very, very mad at Gradle's error prone API: Calling `getCredentials()` on a repository that has no password credentials is what triggers the issue.

It's very nasty because for some reason, it happens on at Gradle sync and never from command line (even if running the `prepareKotlinBuildScriptModel` task).

I'll consider open an issue for Gradle because this is not great.

I'm closing this PR and will submit another PR that actually fixes #227.

LouisCAD:
Bad news: there's no way to workaround this issue without using internal APIs, unless we're okay dropping support for basic authenticated repos or having a way to enable/disable it on demand.

I'll see if I can get Gradle 6.7 to fix that issue to avoid disabling that feature.
Here's the bug report for reference: https://github.com/gradle/gradle/issues/14694

Issue #226, opened 2020-08-22
Replace outdated samples by new ones using the latest version


LouisCAD:
Namely, `sample-android` and `sample-bleeding-egde` haven't been updated in a long time and there's been many changes since then.

We need a single new sample for Android that showcases use of buildscript dependencies, plugins and BoM usage.

jmfayard:
2af05a7 is now merged

PR #233, opened 2020-08-28
226-samples

jmfayard:
#226 upgrade android-sample

- Gradle 6
- Latest plugin usage
- BOM example with Firebase
- Buildscript example
- Plugin example with gradle.site

 #226 remove sample-bleeding-edge

It was there for when the updater was experimental and the rest of the project was not upgraded to Gradle 6. Does not make sense anymore


sample-android/versions.properties
```diff
@@ -1,30 +1,47 @@
...
+## suppress inspection "SpellCheckingInspection" for whole file
+## suppress inspection "UnusedProperty" for whole file
+##
+## Dependencies and Plugin versions with their available updates
+## Generated by $ ./gradlew refreshVersions
+## Please, don't put extra comments in that file yet, keeping them is not supported yet.
+
+plugin.android=4.0.1
+
 plugin.com.louiscad.splitties=0.1.2
```
LouisCAD, line 10: We can remove this now.

```suggestion
```

sample-android/app/build.gradle.kts
```diff
@@ -1,15 +1,11 @@
...
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
     id("com.android.application")
-    id("com.louiscad.splitties")
     id("org.jetbrains.kotlin.android")
     id("org.jetbrains.kotlin.kapt")
-    id("kotlin-android-extensions")
+    id("org.jetbrains.kotlin.android.extensions")
+    id("gradle.site")
```
LouisCAD, line 8: Did you add the gradle.site plugin to provide an example of using a plugin with its id?

sample-android/app/build.gradle.kts
```diff
@@ -19,19 +15,23 @@ tasks.withType<KotlinCompile> {
 }
 
 dependencies {
-    implementation(Kotlin.stdlibJdk7)
+    implementation(platform(Firebase.bom))
+    implementation(Firebase.authentication)
+    implementation("com.google.firebase:firebase-analytics")
```
LouisCAD, line 20: Maybe we should add this one in the built-in dependencies constants?

sample-android/app/build.gradle.kts
jmfayard: Exactly

sample-android/app/build.gradle.kts
jmfayard: it's already there. 
Since this platform constraint thing is new and not well-known, I wanted to show explicitely that there is no version

PR #237, opened 2020-08-29
✚ Add type safety to dependencies.implementation()

jmfayard:
This is a follow up on https://github.com/jmfayard/refreshVersions/pull/229

It focuses on typesafety only

There is an elegant Kotlin solution for every problem.

The IDE now imports two version of `implementation`. 

```
fun DependencyHandler.`implementation`(dependencyNotation: Any): Dependency? =
    add("implementation", dependencyNotation)

@kotlin.Deprecated(ErrorIsNotADependency)
fun DependencyHandler.`implementation`(dependencyNotation: IsNotADependency): Dependency? =
    null
```

Since the second one is more specific, it takes precedence when the parameter is `IsNotADependency` or in practice `DependencyNotationAndGroup`

This way the IDE shows you that made a mistake

GraySky____IdeaProjects_gray-sky-weather__-_build_gradle_kts___app_

### DIFF
--- a/plugins/dependencies/src/main/kotlin/dependencies/DependencyNotationAndGroup.kt
+++ b/plugins/dependencies/src/main/kotlin/dependencies/DependencyNotationAndGroup.kt
@@ -1,11 +1,9 @@
 package dependencies
 
-import org.gradle.kotlin.dsl.IsNotADependency
-
 abstract class DependencyNotationAndGroup(
     group: String,
     name: String
-) : CharSequence, IsNotADependency {
+) : CharSequence {
 
     protected val artifactPrefix = "$group:$name"
 
--- a/plugins/dependencies/src/main/kotlin/dependencies/DependencyNotationAndGroup.kt
+++ b/plugins/dependencies/src/main/kotlin/dependencies/DependencyNotationAndGroup.kt
@@ -1,9 +1,11 @@
 package dependencies
 
+import org.gradle.kotlin.dsl.IsNotADependency
+
 abstract class DependencyNotationAndGroup(
     group: String,
     name: String
-) : CharSequence {
+) : CharSequence, IsNotADependency {
 
     protected val artifactPrefix = "$group:$name"
 
--- a/plugins/dependencies/src/main/kotlin/org/gradle/kotlin/dsl/IsNotADependency.kt
+++ b/plugins/dependencies/src/main/kotlin/org/gradle/kotlin/dsl/IsNotADependency.kt
@@ -6,24 +6,30 @@ import org.gradle.api.artifacts.Dependency
 import org.gradle.api.artifacts.dsl.DependencyHandler
 
 /**
- * Marker interface that prevents the user to add a wrong dependency like
+ * Marker interface that prevents the user to add a wrong dependency
+ * so that the IDE can show a warning
  *
+ * ```
+ * dependencies {
  *     implementation(Testing.kotest)
- *                   ^^^^^^^^^^^^^^^
- *                   IDE: This is not a valid dependency notation but a group of dependency notations
+ *                   ^^^^^^^^^^^^^^
+ *        This is not a valid dependency notation but a group of dependency notations
+ * }
+ * ```
  */
 interface IsNotADependency
 
-const val ErrorIsNotADependency = """This is not a valid dependency notation but a group of dependency notations"""
+private const val ErrorIsNotADependency =
+    """This is not a valid dependency notation but a group of dependency notations"""
 
-@kotlin.Deprecated(ErrorIsNotADependency)
+@Deprecated(ErrorIsNotADependency)
 fun DependencyHandler.`implementation`(dependencyNotation: IsNotADependency): Dependency? =
     null
 
-@kotlin.Deprecated(ErrorIsNotADependency)
+@Deprecated(ErrorIsNotADependency)
 fun DependencyHandler.`api`(dependencyNotation: IsNotADependency): Dependency? =
     null
 
-@kotlin.Deprecated(ErrorIsNotADependency)
+@Deprecated(ErrorIsNotADependency)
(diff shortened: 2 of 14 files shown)


plugins/dependencies/src/main/kotlin/dependencies/DependencyNotationAndGroup.kt
```diff
@@ -1,9 +1,11 @@
 package dependencies
 
+import org.gradle.kotlin.dsl.IsNotADependency
+
 abstract class DependencyNotationAndGroup(
     group: String,
     name: String
-) : CharSequence {
+) : CharSequence, IsNotADependency {
```
LouisCAD, line 8: Actually, `DependencyNotationAndGroup` subclasses are a dependency, that's why they extend `CharSequence`.

LouisCAD (changes requested):

plugins/dependencies/src/main/kotlin/org/gradle/kotlin/dsl/IsNotADependency.kt
```diff
@@ -5,6 +5,11 @@ package org.gradle.kotlin.dsl
 import org.gradle.api.artifacts.Dependency
 import org.gradle.api.artifacts.dsl.DependencyHandler
 
+/**
+ * Marker interface that prevents the user to add a wrong dependency like
+ *     implementation(AndroidX.core)
+ * Since AndroidX.core IsNotADependency, the IDE will show a warning
```
LouisCAD, line 11: `AndroidX.core` is a dependency, we need to use another example.

LouisCAD (changes requested):

LouisCAD:
Something that would be cool it to have a test to ensure we don't forget to extend either `IsNotADependency` or `DependencyNotationAndGroup` for dependency objects. We can do it in another PR though.

plugins/dependencies/src/main/kotlin/org/gradle/kotlin/dsl/IsNotADependency.kt
jmfayard: I updated the example

plugins/dependencies/src/main/kotlin/dependencies/DependencyNotationAndGroup.kt
jmfayard: Ah ah! Now I get it

plugins/dependencies/src/main/kotlin/org/gradle/kotlin/dsl/IsNotADependency.kt
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Marker interface that prevents the user to add a wrong dependency like
+ *
+ *     implementation(Testing.kotest)
+ *                   ^^^^^^^^^^^^^^^
+ *                   IDE: This is not a valid dependency notation but a group of dependency notations
+ */
+interface IsNotADependency
+
+const val ErrorIsNotADependency = """This is not a valid dependency notation but a group of dependency notations"""
```
LouisCAD, line 17: This should be private.
```suggestion
private const val ErrorIsNotADependency = """This is not a valid dependency notation but a group of dependency notations"""
```

plugins/dependencies/src/main/kotlin/org/gradle/kotlin/dsl/IsNotADependency.kt
```diff
@@ -0,0 +1,29 @@
...
+
+@kotlin.Deprecated(ErrorIsNotADependency)
```
LouisCAD, line 19: Why is there the `kotlin.` prefix? Stands for other occurrences in this file as well.

plugins/dependencies/src/main/kotlin/org/gradle/kotlin/dsl/IsNotADependency.kt
```diff
@@ -0,0 +1,29 @@
...
+fun DependencyHandler.`implementation`(dependencyNotation: IsNotADependency): Dependency? =
+    null
+
+@kotlin.Deprecated(ErrorIsNotADependency)
+fun DependencyHandler.`api`(dependencyNotation: IsNotADependency): Dependency? =
+    null
+
+@kotlin.Deprecated(ErrorIsNotADependency)
+fun DependencyHandler.`testImplementation`(dependencyNotation: IsNotADependency): Dependency? =
+    null
```
LouisCAD, line 29: We can add other configurations (`compileOnly`, `runtimeOnly`& `classpath`), plus their overloads taking a lambda (though it's not fully supported for rich `version { … }` by refreshVersions yet).

LouisCAD (changes requested):

plugins/dependencies/src/main/kotlin/org/gradle/kotlin/dsl/IsNotADependency.kt
jmfayard: fixed

plugins/dependencies/src/main/kotlin/org/gradle/kotlin/dsl/IsNotADependency.kt
jmfayard: done, thanks

plugins/dependencies/src/main/kotlin/org/gradle/kotlin/dsl/IsNotADependency.kt
jmfayard: Le mieux est l'ennemi du bien :)

LouisCAD: @jmfayard Why not add it, and why would it be the enemy of good in this case?

Issue #235, opened 2020-08-29
$ ./gradlew buildSrcVersions generating Libs.kt


jmfayard:
*Update*: Pull request open https://github.com/jmfayard/refreshVersions/pull/245

**Context**

`buildSrcVersions` had two features

- `buildSrc/.../Versions.kt` now replaced by `versions.properties` and the plugin `refreshVersions-core`
- `buildSrc/.../Libs.kt` which provided auto-completion of the dependency notations in all Gradle modules for both the Kotlin and the Groovy DSL. 

The plugin `de.fayard.refreshVersions` now contains dependencies constants for common libraries, which is a better solution especially for single-module Gradle projects using the Kotlin DSL. 

But the `buildSrc/.../Libs.kt` alternative is still a good one for multi-modules projects, especially if they are large (we can't have all the libraries in mavenCentral) or of they use the Groovy DSL - for the auto-completion currently does not work there and it would be a while until they fix Android Studio and IntelliJ. 

**Describe the solution you'd like**

Proposal is to do a resurrection of `buildSrcVersions` in its version 0.7.0, keeping just `Libs.kt`, using the place holder version `_` instead of `Versions.kt` 

https://github.com/jmfayard/refreshVersions/tree/v0.7.0/plugin

 
**Describe alternatives you've considered**

The task should probably be called `:buildSrcLibs` with an alias `:buildSrcVersions` to not disturb the habits.

In which Gradle plugin should the code be? In the `de.fayard.refreshVersions` plugin or in the existing `de.fayard.buildSrcVersions` plugin? Latter is cleaner, former adds more configuration to do. Should the plugin versions be synchronized with refreshVersions or independant?

jmfayard:
Corriger l#'ordre dans lequel les properties sont générées

rvp-diconium:
The current `refreshVersions` during the migration does a clear separation on implementation/testImplementation/etc  
I believe it should be possible then to bring the same organization to `buildSrcVersions`. Example:
```
object Libs {
    const val appcompat: String = "androidx.appcompat:appcompat:_"

    // ..
}
object TestLibs {
    const val junit: String = "...junit..."

    // ..
}
object AndroidTestLibs {
    const val espresso: String = "...espresso..."

    // ..
}
```

What do you think?

jmfayard:
I don't think so

Refresh versions are crafted manually for a limited set of dependencies
white buildsrcversions generates automatically whatever dependencies you
throw at him.

[...]

PR #245, opened 2020-09-06
$ ./gradlew buildSrcLibs generating Libs.kt

jmfayard:
WIP for https://github.com/jmfayard/refreshVersions/issues/235

Roadmap:

- [x] Brings back the task `./gradle buildSrcVersions` but use the version placehoder `_`  instead of using `Versions.kt`
- [x] Create an alias `buildSrcLibs`
- [ ] Add missing keys in `versions.properties`, like in branch https://github.com/jmfayard/refreshVersions/tree/poc-all-versions
- [x] Delete the file `buildSrc/src/main/kotlin/Versions.kt` if it exists
- [ ] Write the documentation for the website
- [ ] Do a release and finish https://github.com/http4k/http4k/pull/499


jmfayard:
`$ gradle buildSrcVersions` This now works on the `sample-multi-modules` and on http4k https://github.com/http4k/http4k/pull/499

LouisCAD:
Wouldn't it be better to name it `buildSrcLibs` since the versions are no longer in the buildSrc? An alias can still be kept for people migrating from `buildSrcVersions`, and we can add a warning to have them use the new name if we want.

jmfayard:
Yes, it's on my todo-list, the task will be called `buildSrcLibs` but `buildSrcVersions` will do the same thing for those used to that name

LouisCAD:
Alright, I'm renaming this PR so we start getting used to the new name 🙂

Issue #250, opened 2020-09-14
Add Material Components Compose theme adapter dependency constant


LouisCAD:
"MDC-Android Compose Theme Adapter"

https://github.com/material-components/material-components-android-compose-theme-adapter

jmfayard:
Done

Issue #238, opened 2020-08-29
Suggest dependencies to add


jmfayard:
Want to contribute new libraries?

Propose it here.

We have homework on our side, we wants to make it easier to contribute a new group: 

Proposal for simplifying adding a dependency group https://github.com/jmfayard/refreshVersions/pull/262

- [ ] Add Kodein https://github.com/jmfayard/refreshVersions/pull/253/files
- [x] Add Http4k https://github.com/jmfayard/refreshVersions/pull/253/files
- [ ] Add Koin https://github.com/jmfayard/refreshVersions/pull/253/files
- [x] Add new kotest modules https://github.com/jmfayard/refreshVersions/pull/261
- [x] Add Spring dependencies https://github.com/jmfayard/refreshVersions/issues/238#issuecomment-685304844
- [ ] See comment from simonsickle https://github.com/jmfayard/refreshVersions/issues/238#issuecomment-709623195
- [ ] Dependencies from the Kotlin libraries Playground https://github.com/jmfayard/refreshVersions/issues/238#issuecomment-709785409
- [ ] Add GraphQL https://github.com/jmfayard/refreshVersions/issues/238#issuecomment-711157530
- [x] Add Http4k https://github.com/jmfayard/refreshVersions/pull/253/files

LouisCAD:
I'm not sure we need to add Guice, nor Kotlin unfriendly Gson (lying nullability).

NikkyAI:
those non-kotliny things would fit well enough into `buildSr`c for people who do end up using them a lot (or a includeBuild/plugin)

jmfayard:
RxJava is also missing

jmfayard:
Also for backend people we already have ktor and we should add the dependencies for the Spring framework that are displayed on https://start.spring.io/

Using the web interface, I added all available dependencies which generated the following Gradle file

```
import org.jetbrains.kotlin.gradle.tasks.KotlinCompile

plugins {
	id("org.springframework.boot") version "2.3.3.RELEASE"
	id("io.spring.dependency-management") version "1.0.10.RELEASE"
	id("org.asciidoctor.convert") version "1.5.8"
	kotlin("jvm") version "1.3.72"
	kotlin("plugin.spring") version "1.3.72"
	kotlin("plugin.jpa") version "1.3.72"
}

group = "com.example"
version = "0.0.1-SNAPSHOT"
java.sourceCompatibility = JavaVersion.VERSION_11

configurations {
	compileOnly {
[...]

PR #253, opened 2020-09-26
dependencies-add

jmfayard:
- Add Material Components Compose theme adapter dependency constant #250
- Add Koin, Http4k, Kodein #238


jmfayard:
What kind of rules should I add? I don't know how this works.
Also it would be cool to have a checklist in `CONTRIBUTING.md` of how to contribute new dependencies like this @LouisCAD

jmfayard:
There is also a rule missing for *spek*
I had to do it manually in the playground-repo

```
// versions.properties
version.org.spekframework.spek2..spek-dsl-jvm=version.spek

version.org.spekframework.spek2..spek-runner-junit5=version.spek

version.org.spekframework.spek2..spek-runtime-jvm=version.spek

version.spek=2.0.13
```

plugins/dependencies/src/main/kotlin/dependencies/Http4k.kt
```diff
@@ -0,0 +1,190 @@
...
+
+    /**
+     * [JSON RPC](https://www.http4k.org/guide/modules/jsonrpc/)
+     */
+    const val jsonrpc = "$prefix-jsonrpc"
+
+    /**
+     * [Metrics](https://www.http4k.org/guide/modules/metrics/)
+     */
+    const val metrics_micrometer = "$prefix-metrics-micrometer"
```
LouisCAD, line 49: Why do we have snake_case here while we have camelCase elsewhere?

LouisCAD:
@jmfayard Could you split this PR per library family? That will make review easier and will enable quicker integration for the libraries I'm familiar with.

PR #255, opened 2020-09-30
Add Google ML Kit and deprecate dependency notations of deprecated Firebase ML Kit artifacts

LouisCAD:


plugins/dependencies/src/main/resources/refreshVersions-rules/google-version-alias-rules.txt
```diff
@@ -24,3 +24,6 @@ com.google.dagger:hilt-android(-*)
 
 com.google.androidbrowserhelper:androidbrowserhelper
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+com.google.mlkit:*
+    ^^^^^^^^^^^^.^
```
LouisCAD, line 29: @jmfayard You can take note of that addition. Might help you if you have to add further dependency notations.

I'll document that notation at some point (and fix the few edge cases I'm aware of where error reporting in the syntax is incomplete)

PR #258, opened 2020-10-07
Add ApolloGraphQL

martinbonnin:
Add ApolloGraphQL to the dependencies: https://github.com/apollographql/apollo-android


LouisCAD:
Thanks for the contribution @martinbonnin!
I need to update the contributing guide (currently WIP) so you get the proper instructions.
I'll do so, notify you when it's published, and I'll review this before the 0.9.7 release (unless I need to publish an important fix before).

jmfayard:
I added your change to the issue https://github.com/jmfayard/refreshVersions/issues/238
When we are ready, we will integrate it

PR #262, opened 2020-10-13
Proposal for simplifying adding a dependency group

jmfayard:
To be discussed by phone


plugins/dependencies/src/main/kotlin/dependencies/Kodein.kt
```diff
@@ -0,0 +1,68 @@
...
+        val ktor = module("kodein-di-framework-ktor-server-jvm")
+        val tornadofx = module("kodein-di-framework-tornadofx-jvm")
+    }
+}
+
+
+open class DependencyGroup(
+    val group: String,
+    val rule: Rule? = null,
+    val usePlatformConstraints: Boolean = false
```
LouisCAD, line 39: We'd want to give users the choice, just like we do for Firebase BoM where you can access the `no-BoM` variant.

jmfayard: Not a problem, see next commit

plugins/dependencies/src/main/kotlin/dependencies/Kodein.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * painless Kotlin dependency injection
+     *
+     * - [Official website here](https://kodein.org/di/)
+     * - GitHub page: [Kodein-Framework/Kodein-DI](https://github.com/Kodein-Framework/Kodein-DI)
+     * - [GitHub Releases here](https://github.com/Kodein-Framework/Kodein-DI/releases)
+     */
+    val di = DI(usePlatformConstraints = true)
+    val `di-no-bom` = DI(usePlatformConstraints = false)
```
LouisCAD, line 18: This is one of the multiple ways to organise/namespace the no BoM version (and this one is inconsistent with what we have in `Firebase`).

We'll have to discuss this and settle on a convention that we can then mention in the contributing guide.

LouisCAD:
Gives food for thoughts, putting this into my priority list.

LouisCAD:
See issue #275

PR #266, opened 2020-10-15
Fix room coroutine version

raychenon:
AndroidX.room.coroutines version is not 2.2.5 (group) but  version 2.1.0-alpha04


LouisCAD:
Hi, that's not how it works. YOU define the version in YOUR project. refreshVersions doesn't hardcode the versions.

raychenon:
Hi @LouisCAD ,
I was discussing with @jmfayard  yesterday about a different version for an artifact than  its groupId.
He suggested to replace the placeholder_ by the version directly.
Or we can create a different object `RoomAlpha: IsNotADependency` for a different version.

LouisCAD:
Ok, it's because the artifact has been abandoned in favor of Room KTX. I'm deprecating it from refreshVersions for the impending release 0.9.7, and we'll remove it in a future release.

raychenon:
Correct, 
shouldn't we just remove this artifact instead of deprecating it ? When you use it, the version is not correct.
Please let me delete this line :)

LouisCAD:
I didn't want to break your build further, that's why I only deprecated it for now.

We have a plan to improve deprecation and removals handling that would not break your builds, but it needs some work.

raychenon:
OK, it makes sense for `removals handling that would not break your builds`

PR #274, opened 2020-10-16
✚ We support Gradle >= 6.3

jmfayard:
In Gradle 6.2, the task refreshVersions fail with this error:

org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':refreshVersions'.
Caused by: java.lang.NoClassDefFoundError: kotlin/coroutines/AbstractCoroutineContextKey

https://gradle.com/s/shp7hbtd3i3ii

How I tested the thing:

```
$ echo "Install SDK man"
$ open https://sdkman.io/
$ sdk install gradle 6.2
$ sdk use gradle 6.2
$ cd sample-kotlin
$ gradle help
FAILURE: Build failed with an exception.

* Where:
Settings file '/Users/jmfayard/IdeaProjects/refreshVersions/sample-kotlin/settings.gradle.kts' line: 28

* What went wrong:
The plugin "de.fayard.refreshVersions" only works with Gradle 6.3 and above.
```

### DIFF
--- a/plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/RefreshVersionsCoreSetup.kt
+++ b/plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/RefreshVersionsCoreSetup.kt
@@ -8,6 +8,8 @@ import de.fayard.refreshVersions.core.internal.resolveVersion
 import de.fayard.refreshVersions.core.internal.setupVersionPlaceholdersResolving
 import org.gradle.api.initialization.Settings
 import org.gradle.kotlin.dsl.apply
+import org.gradle.tooling.UnsupportedVersionException
+import org.gradle.util.GradleVersion
 import java.io.File
 
 /**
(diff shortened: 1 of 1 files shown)


plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/RefreshVersionsCoreSetup.kt
```diff
@@ -42,6 +44,10 @@ fun Settings.bootstrapRefreshVersionsCore(
     artifactVersionKeyRules: List<String> = emptyList(),
     versionsPropertiesFile: File = rootDir.resolve("versions.properties")
 ) {
+    val supportedGradleVersion = "6.3" // 6.2 fail with this error: https://gradle.com/s/shp7hbtd3i3ii
+    if (GradleVersion.current() < GradleVersion.version(supportedGradleVersion)) {
+        throw UnsupportedVersionException("""The plugin "de.fayard.refreshVersions" only works with Gradle $supportedGradleVersion and above.""")
```
LouisCAD, line 49: Should we include some guidance on how to upgrade or do we let the users figure it out on their own (which is also acceptable)?

jmfayard: If and when we have a page on the website on how to upgrade to Gradle 6, we can put a link there. 
For now I wouldn't care.

LouisCAD: @jmfayard We can simply link the Gradle migration docs:
- From Gradle 6+: https://docs.gradle.org/current/userguide/upgrading_version_6.html
- From Gradle 5.x: https://docs.gradle.org/current/userguide/upgrading_version_5.html
- From Gradle 4.x: https://docs.gradle.org/current/userguide/upgrading_version_4.html

jmfayard:
Thanks for the URL
I want to put them on the website, so that we can give more details and update as often as we want.
But since we didn't agree yet on the name of the GitHub organization, the URL will change, so it doesn't make sense at the moment to add it.

LouisCAD:
Putting the Gradle URLs is good enough for now, and better than nothing, right?

PR #278, opened 2020-10-19
./gradlew buildSrcLibs

### LINKED ISSUE #235: $ ./gradlew buildSrcVersions generating Libs.kt

jmfayard:
Closes #235 

WIP for https://github.com/jmfayard/refreshVersions/issues/235

Roadmap:

- [x] Brings back the task `./gradle buildSrcVersions` but use the version placehoder `_`  instead of using `Versions.kt`
- [x] Create an alias `buildSrcLibs`
- [x] Add missing keys in `versions.properties`, like in branch https://github.com/jmfayard/refreshVersions/tree/poc-all-versions
- [x] Delete the file `buildSrc/src/main/kotlin/Versions.kt` if it exists
- [ ] Write the documentation for the website
- [ ] Do a release and finish https://github.com/http4k/http4k/pull/499


LouisCAD:
@jmfayard Shouldn't this target the `develop` branch instead of the `master` branch?

jmfayard:
Indeed, facepalm!

LouisCAD:
I think you can just re-target the branch.

Issue #241, opened 2020-08-31
Setup refreshVersions via a Plugin<Settings>


jmfayard:
NikkyAi asked on Slack whether instead of messing with the buildscript classpath, we could do something similar to the Gradle entreprise plugin

```kotlin
// settings.gradle.kts
plugins {
   id("de.fayard.refreshVersions.settings") version "0.9.XXXX"
}

refreshVersions {

}
```

- this would make the setup cleaner - no more messing with the classpath
- we could add configuration options via a Gradle extension instead of reinventing the wheel https://github.com/jmfayard/refreshVersions/issues/202

jmfayard:
I've done an experiment in the `plugin-settings` branch and it seems to me that it would work!

https://github.com/jmfayard/refreshVersions/compare/develop...plugin-settings?expand=1

LouisCAD:
We need to check if we can support the configuration that the bootstrap currently allows, and see if we can get it early enough for it to work for `buildscript`.

If yes, then we might be able to support both ways.

LouisCAD:
Implemented in #280 and now merged into the develop branch.

LouisCAD:
Finally released in 0.10.0

PR #280, opened 2020-10-21
$ ./gradlew buildSrcLibs

### LINKED ISSUE #235: $ ./gradlew buildSrcVersions generating Libs.kt

jmfayard:
Closes #235 

Now with documentation


jmfayard:
image

jmfayard:
Use the API from https://github.com/jmfayard/refreshVersions/pull/281 once it's merged

docs/setup.md
```diff
@@ -35,10 +35,12 @@ See the example snippet below:
...
 import com.example.something // Imports at the top, as usual.
 
+pluginManagement {} // Optional
+
 buildscript {
    // We will setup refreshVersions here, see below.
 }
-pluginManagement {} // Optional
+
 plugins {} // Optional
```
LouisCAD, line 44: Are you sure this is the correct order? Why not put this change in another PR? It's not really related to buildSrcLibs.

docs/setup.md
```diff
@@ -77,6 +79,27 @@ Here is how to configure gradle refreshVersions:
...
 
 
+### If you upgrade from the plugin buildSrcVersions
+
+Before refreshVersions, [there was the plugin buildSrcVersions](https://dev.to/jmfayard/better-dependency-management-in-android-studio-3-5-with-gradle-buildsrcversions-34e9)
+
+If your project is using it, remove all its configuration from the top `build.gradle[.kts]` file
+
+```diff
+// build.gradle[.kts]
```
LouisCAD, line 89:
```suggestion
=== "build.gradle.kts"
```diff
```

This should show nice file name snippet header as we have in the rest of the setup file (please check locally).

mkdocs.yml
```diff
@@ -64,6 +64,7 @@ nav:
   - 'Setup': setup.md
   - 'Add Dependencies': add-dependencies.md
   - 'Update Dependencies': update-dependencies.md
+  - 'Gradle buildSrcVersions': gradle-buildsrcversions.md
```
LouisCAD, line 67: Shouldn't we name if `buildSrcLibs` here?

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/internal/PluginConfig.kt
```diff
@@ -0,0 +1,79 @@
+package de.fayard.internal
+
+
+@Suppress("unused")
+object PluginConfig {
```
LouisCAD, line 5: Why is this public instead of `internal`?

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/internal/KotlinPoetry.kt
```diff
@@ -0,0 +1,106 @@
...
+fun constStringProperty(name: String, initializer: CodeBlock, kdoc: CodeBlock? = null) =
+    PropertySpec.builder(name, String::class)
+        .addModifiers(KModifier.CONST)
+        .initializer(initializer)
+        .apply {
+            if (kdoc != null) addKdoc(kdoc)
+        }.build()
+
+
+fun constStringProperty(name: String, initializer: String, kdoc: CodeBlock? = null) =
```
LouisCAD, line 103: Why is this all public instead of `internal`?

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/BuildSrcLibsTask.kt
```diff
@@ -0,0 +1,141 @@
...
+        val deps = allDependencies.checkModeAndNames(resolvedUseFqdn)
+
+        val libsFile: FileSpec = kotlinpoet(deps)
+
+        libsFile.writeTo(outputDir)
+        OutputFile.logFileWasModified(OutputFile.LIBS.path, OutputFile.LIBS.existed)
+    }
+
+
+    fun findDependencies(): List<Dependency> {
```
LouisCAD, line 83: Shouldn't this one be `internal`?

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/VersionsPropertiesWriting.kt
```diff
@@ -67,6 +67,21 @@ internal fun writeWithAddedVersions(
     versionsFile.writeText(newFileContent)
 }
 
+fun writeWithNewEntries(versionsFile: File, newEntries: List<Pair<String, String>>) {
```
LouisCAD, line 70: This one should be `internal`.

LouisCAD (changes requested):
This is a partial review. I'll will come back at it once I'm done with #281.

I'll also need to review the website updates in depth.

docs/setup.md
jmfayard: Yes I'm sure, here is what happens if I use the documented order:

```
$ ./gradlew help
> /Users/jmfayard/IdeaProjects/refreshVersions/sample-kotlin/settings.gradle.kts:3:1: 
Unexpected `buildscript` block found. `buildscript` can not appear before `pluginManagement`.
```

mkdocs.yml
jmfayard: I use buildSrcVersions here on purpose.
From a SEO perspective, we want people who already know buildSrcVersions too google exactly that and land on our page that explains by what it was replaced instead of my old medium articles describing buildSrcVersions

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/internal/PluginConfig.kt
jmfayard: Good points, I will take care of it once https://github.com/jmfayard/refreshVersions/pull/281 is merged

docs/setup.md
jmfayard: Nice

jmfayard:
@LouisCAD I'm done, buildSrcLibs now use the VersionsPropertiesModel and works on the `sample-multi-modules` as well as in the http4k project and also in the playground at this commit https://github.com/LouisCAD/kotlin-libraries-playground/commit/1586afe8579e12e2fa63114da877198dd38a22d1

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/versions/VersionsPropertiesWriting.kt
```diff
@@ -10,8 +10,18 @@ import de.fayard.refreshVersions.core.internal.isAVersionAlias
 import de.fayard.refreshVersions.core.internal.versions.VersionsPropertiesModel.Companion.availableComment
 import de.fayard.refreshVersions.core.internal.versions.VersionsPropertiesModel.Section.Comment
 import de.fayard.refreshVersions.core.internal.versions.VersionsPropertiesModel.Section.VersionEntry
+import org.gradle.api.artifacts.ExternalDependency
 import java.io.File
 
+fun writeMissingEntriesInVersionProperties(newEntries: Map<String, ExternalDependency>) {
```
LouisCAD, line 16: Any reason this one is not `internal` @jmfayard?

jmfayard: yes, it won't compile if it's internal. 
Feel free to improve it :)

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/BuildSrcLibsTask.kt
```diff
@@ -100,39 +100,38 @@ open class BuildSrcLibsTask : DefaultTask() {
...
-    val versionsProperties = RefreshVersionsConfigHolder.readVersionProperties()
+    val versionsProperties = RefreshVersionsConfigHolder.readVersionsMap()
     val projectsWithHardcodedDependenciesVersions: List<Project> = rootProject.allprojects.filter {
         it.countDependenciesWithHardcodedVersions(versionsProperties) > 0
     }
 
     return projectsWithHardcodedDependenciesVersions.flatMap { project ->
         project.configurations.filterNot { configuration ->
-            configuration.shouldBeIgnored() || 0 == configuration.countDependenciesWithHardcodedVersions(versionsProperties, versionKeyReader)
+            configuration.shouldBeIgnored() || 0 == configuration.countDependenciesWithHardcodedVersions(versionsProperties, RefreshVersionsConfigHolder.versionKeyReader)
```
LouisCAD, line 110: Would be better to put these arguments on separate lines for readability I think.

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/versions/VersionsPropertiesWriting.kt
LouisCAD: Then it should be annotated with `@InternalRefreshVersionsApi`. Ideally, it'd also not be top-level, but we already have a bunch of top level symbols in this case, so I'd revisit that latter point later.

LouisCAD:
Regarding the documentation, I think it'd be better to separate the part for current buildSrcVersions users from the part for newcomers to buildSrcLibs.

Also, we need to keep in mind that I'll manage to workaround the issue we have with Groovy DSL and bundled dependency notations as part of #241, so using `buildSrcLibs` will not be a necessity in the near future.

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/BuildSrcLibsTask.kt
```diff
@@ -0,0 +1,139 @@
...
+}
+
+
+internal fun findMissingEntries(configurations: List<Pair<Project, Configuration>>): Map<String, ExternalDependency> {
+
+    val versionsProperties = RefreshVersionsConfigHolder.readVersionsMap()
+
+    val versionKeyReader = RefreshVersionsConfigHolder.versionKeyReader
+
+    val dependencyMap = configurations.flatMap { (project, configuration) ->
```
LouisCAD, line 124: `project` is never used here. Why put it in the configurations list in the first place since it's not used at all?

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/BuildSrcLibsTask.kt
```diff
@@ -0,0 +1,139 @@
...
+
+    private fun findDependencies(): List<Dependency> {
+        val allDependencies = mutableListOf<Dependency>()
+        project.allprojects {
+            val projectName = name
+            // println("Configurations: " + configurations.map { it.name })
+            allDependencies += (configurations + buildscript.configurations)
+                //.filter { it.name in PluginConfig.knownConfigurations }
+                .flatMap {
+                    val configurationName = "$projectName:${it.name}"
```
LouisCAD, line 88: This is not used. Why create that useless string?

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/BuildSrcLibsTask.kt
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+
+    private fun findDependencies(): List<Dependency> {
+        val allDependencies = mutableListOf<Dependency>()
+        project.allprojects {
+            val projectName = name
+            // println("Configurations: " + configurations.map { it.name })
+            allDependencies += (configurations + buildscript.configurations)
+                //.filter { it.name in PluginConfig.knownConfigurations }
```
LouisCAD, line 86: Why this commented out code with no explanation?

If there's things to revisit or shortcuts taken for some reason, please put the full comments in the code for future anyone contributing to the codebase.

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/internal/OutputFile.kt
```diff
@@ -0,0 +1,44 @@
+package de.fayard.refreshVersions.internal
+
+import org.gradle.api.Project
+
+internal enum class OutputFile(var path: String, var existed: Boolean = false, val alternativePath: String? = null) {
+    OUTPUTDIR("buildSrc/src/main/kotlin"),
```
LouisCAD, line 6: Would probably be better renamed to `OUTPUT_DIR`.

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/internal/OutputFile.kt
```diff
@@ -0,0 +1,44 @@
...
+    fun logFileWasModified(delete: Boolean = false) {
+        logFileWasModified(path, existed, delete)
+    }
+
+    companion object {
+
+        fun logFileWasModified(path: String, existed: Boolean, delete: Boolean = false) {
+            val ANSI_RESET = "\u001B[0m"
+            val ANSI_GREEN = "\u001B[32m"
+            val ANSI_RED = "\u001B[31m"
```
LouisCAD, line 30: Please move these as private constants in that same companion object, below, so the IDE doesn't warn about the case not following the conventions.

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/internal/KotlinPoetry.kt
```diff
@@ -0,0 +1,103 @@
+package de.fayard.refreshVersions.internal
+
+import com.squareup.kotlinpoet.*
+
+
+internal data class Dependency(
+    val group: String = "",
+    val module: String = "",
+    val version: String = ""
+) {
```
LouisCAD, line 10: This can be confused with `Dependency` from Gradle. Do you have other name proposal that would fit for the use you have there?

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/internal/KotlinPoetry.kt
```diff
@@ -0,0 +1,103 @@
...
+        }
+    )
+
+    override fun toString() = groupModuleVersion()
+}
+
+
+internal class Deps(
+    val dependencies: List<Dependency>,
+    val modes: Map<Dependency, VersionMode>,
```
LouisCAD, line 29: This is not used according to the IDE. Why is it there, then?

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/internal/KotlinPoetry.kt
```diff
@@ -0,0 +1,103 @@
...
+                else -> CodeBlock.of("%S", d.groupModuleUnderscore())
+            }
+            constStringProperty(
+                name = deps.names[d]!!,
+                initializer = libValue,
+                kdoc = null
+            )
+        }
+
+    val Libs = TypeSpec.objectBuilder("Libs")
```
LouisCAD, line 59: Not following the Kotlin naming conventions. I think we should name it `libsTypeSpec` or alike.

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/internal/KotlinPoetry.kt
```diff
@@ -0,0 +1,103 @@
...
+        .addKdoc(PluginConfig.KDOC_LIBS)
+        .addProperties(libsProperties)
+        .build()
+
+
+    val LibsFile = FileSpec.builder("", "Libs")
```
LouisCAD, line 65: Likewise, this should be named `libsFileSpec` or alike.

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/internal/KotlinPoetry.kt
```diff
@@ -0,0 +1,103 @@
...
+        .build()
+
+    return LibsFile
+
+}
+
+internal fun List<Dependency>.checkModeAndNames(useFdqnByDefault: List<String>): Deps {
+    val dependencies = this
+
+    val modes: MutableMap<Dependency, VersionMode> = dependencies.associate { d: Dependency ->
```
LouisCAD, line 77: Can be replaced with `associateWith` (there's an IDE intention to do so).

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/internal/KotlinPoetry.kt
```diff
@@ -0,0 +1,103 @@
...
+import com.squareup.kotlinpoet.*
+
+
+internal data class Dependency(
+    val group: String = "",
+    val module: String = "",
+    val version: String = ""
+) {
+    val name: String get() = module
+    fun groupModuleVersion() = "$group:$module:$version"
```
LouisCAD, line 12: Is there any plan to ever use this outside of `toString()`?

LouisCAD (changes requested):

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/internal/PluginConfig.kt
```diff
@@ -0,0 +1,79 @@
...
+        |repositories {
+        |    mavenCentral()
+        |}
+        """.trimMargin()
+
+    internal fun computeUseFqdnFor(
+            dependencies: List<Dependency>,
+            configured: List<String>,
+            byDefault: List<String> = MEANING_LESS_NAMES
+    ): List<String> {
```
LouisCAD, line 64:
```suggestion
    internal fun computeUseFqdnFor(
        dependencies: List<Dependency>,
        configured: List<String>,
        byDefault: List<String> = MEANING_LESS_NAMES
    ): List<String> {
```

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/BuildSrcLibsTask.kt
jmfayard: deleted

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/BuildSrcLibsTask.kt
jmfayard: deleted

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/BuildSrcLibsTask.kt
jmfayard: deleted, it's not needed

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/internal/OutputFile.kt
jmfayard: done

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/internal/KotlinPoetry.kt
jmfayard: deleted

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/internal/KotlinPoetry.kt
jmfayard: done

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/internal/KotlinPoetry.kt
jmfayard: Yes

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/internal/PluginConfig.kt
jmfayard: done

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/internal/KotlinPoetry.kt
jmfayard: done

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/internal/OutputFile.kt
jmfayard: done

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/internal/KotlinPoetry.kt
jmfayard: Renamed to `Library`

jmfayard:
@LouisCAD I have my iMac again!
I handled your comments, and tested that it still works on the http4k project. Any objection to merge?

LouisCAD:
Great you got the machine back!
And great for handling the feedback!

As we agreed on the phone, I'll put this in its own module and restore the `de.fayard.buildSrcVersions` plugin id, along with the new `de.fayard.buildSrcLibs` one, so we don't pull the kotlinpoet dependency and this code when not needed (which can speed up ephemeral CIs like GitHub Actions a bit as well for our users).

Once I'm done with it, I'll still have to complete the review and needed edits for the documentation, and then I'll merge it.

Hopefully, I'll be done by this weekend 🤞

jmfayard:
@LouisCAD I've thought about it and I think it's a bad idea to upgrade the **plugin** `de.fayard.buildSrcVersions` for now. 
- if the plugin was configured with [`plugins { buildSrcVersions }`](https://github.com/http4k/http4k/blob/3ef9a6d2705050de67907ff629c286fbe5f8927f/buildSrc/src/main/kotlin/Versions.kt#L138-L144) the build will fail after I delete `Versions.kt`
- a user who upgrade without setting up refreshVersions - which is what we should assume will happen - will have `versions.properties` generated but not used since refreshVersions is not configured
- the documentation in its current state would be wrong
- also refreshVersions is currently less stable than buildSrcVersions, it's not the right time to ask everyone to upgrade

In summary, we have now something that works, I would rather not change it to something that doesn't work.

LouisCAD:
@jmfayard We can still do it for `buildSrcLib` then, and revisit smooth migration for `buildSrcVersions` users later on.

jmfayard:
I don't understand your comment.

jmfayard:
Since buildSrcLibs and buildSrcVersions are the same thing, the suggestion adresses none of my concerns.
I would rather merge what we have, which works, and revisit later once we have no better things to do, including a smooth migration like you said.

LouisCAD:
The plugin with id `de.fayard.buildSrcLibs` can still be used without breaking existing users of `de.fayard.buildSrcVersions`. We can reuse that latter one once we have a smooth migration path.

So I'd do the split as we agreed earlier, but for `de.fayard.buildSrcLibs` only.

LouisCAD:
I'll work on #241 as part of this PR so we can have settings plugins for all the plugins.

In fact, I'm already working on it.

plugins/buildSrcLibs/src/main/kotlin/de/fayard/buildSrcLibs/BuildSrcLibsPlugin.kt
```diff
@@ -0,0 +1,20 @@
+package de.fayard.buildSrcLibs
+
+import org.gradle.api.Plugin
+import org.gradle.api.initialization.Settings
+import org.gradle.kotlin.dsl.register
+
+class BuildSrcLibsPlugin : Plugin<Settings> {
```
jmfayard, line 7: @LouisCAD is it intentional that you make it a Plugin of Settings instead of a normal plugin?
I don't see why that would be needed

LouisCAD: Yes, it's intentional, but I might make it a `Plugin<Any>` depending on the outcome of my experiments.

Anyway, it's going to apply the refreshVersions plugin automatically.

jmfayard: I have set-up JetPack for development version, but because the plugin is a `Plugin<Settings>`, I don't know how to configure it correctly =>https://github.com/jmfayard/refreshVersions/issues/290

LouisCAD: @jmfayard The generic type parameter is not relevant to the maven publication, it shouldn't be different from another plugin.

If you are asking about getting the maven coordinates from the plugin id, there's an image and details about this on that page: https://docs.gradle.org/current/userguide/plugins.html

jmfayard:
Louis, can you merge and make a note of what is missing ?

LouisCAD:
I need to:
1. Implement settings plugins updates (includes self-updates)
2. Make the buildSrcLibs plugin apply the refreshVersions plugin
3. Update the docs
4. Ensure the bootstrap cannot be run in addition to applying the plugin
5. Deprecate the old bootstrap
6. Ideally automate migration for old bootstrap to settings syntax

I'll do my best to do as many of these steps as possible until tomorrow afternoon and I'll merge this PR.

jmfayard:
I prepared a PR for the kotlin libraries playground to use the new plugin syntax and be pre-configured with jitpack https://github.com/LouisCAD/kotlin-libraries-playground/pull/89

LouisCAD:
Implementing settings plugins updates is tricky to implement with the available Gradle APIs (no way to query the ids of the applies plugins for the plugins applied via an id), but I finally have a path to make it work, including for buildSrc which was posing me a few challenges.

It's very important to have this ready and reliable before releasing a version using the new setup, because our users will expect refreshVersions to let them know it has available updates, so I prefer to not merge this PR just yet and keep working on this until it's done, which should not take more than a few evenings, hopefully less.

I'll keep you updated @jmfayard, I know this PR has been standing here for a while now and I also want it to be merged and move on to other things in this project as well as other projects.

LouisCAD:
I implemented the tricky part of finding settings plugins updates, which is the first step out of two for this important feature.

The second step is editing the Gradle settings files safely. I expect this to be fairly easy to implement since I've already done similar stuff in this project.

jmfayard:
Nice

LouisCAD:
Just FYI, I didn't have much time to work on this:
Writing the tests that cover the different possible edge cases of the required parsing and edition of the Gradle settings files is tricky, so I need deep focus.

I'll dedicate some time for that this weekend.
If you have time, we can also do a Code With Me session to write these test data @jmfayard .

LouisCAD:
There's still [work to do](https://github.com/jmfayard/refreshVersions/pull/280#issuecomment-727254984), but as we agreed on the phone, we're merging this into `develop` and will complete it as separate pull requests before the next non dev release.

LouisCAD (commented):
Late review: Good enough. 😄

jmfayard:
It's merged 🤩🤩🤩

Issue #134, opened 2019-11-15
New updater checklist


LouisCAD:
- [x] Serialize the repositories from `buildSrc` and `Settings` to use them to get versions.
- [x] Automatically use the most stable version if a placeholder isn't resolved.
- [x] Update or remove the Gradle extension so there's no longer ignored configuration.
- [x] Add empty line after each version property
- [x] Make `maven-metadata.xml` files download parallel (coroutines + `Dispatchers.IO`?).
- [x] Make parsing of the files parallel (coroutines + `Dispatchers.Default`?).
- [x] Report hardcoded & dynamic versions (#126)
- [x] New versions sorting strategy #153 
- [x] Show user-friendly error messages when unable to find available versions of an artifact from the given repositories.
- [x] Ease adding dependencies (#125)
- [x] Support cancellation for in-flight network requests.
- [x] Support update of the plugin itself
- [x] Support advertising Gradle update, link to migration guide and write the commands to update the wrapper ([API here](https://services.gradle.org/versions/)).
- [x] Keep user comments added above a property
- [ ] Specify behavior for unused version entries (remove as before by default? add unused metadata comment?)
- [ ] Allow setting minimum stability level for version property (via a comment above it)
- [ ] Allow setting minimum version, via a comment above it (use-case: waiting for that bugfix release to upgrade)
- [ ] Allow overriding filtering & sorting mode per version property (via a comment above it)
- [ ] Allow disabling updates for a version property (via a comment above it)

jmfayard:
Can be considered as done, we have now a way to configure the plugin via `settings { } `

LouisCAD:
There's items left in the checklist, so the issue overall is still relevant and is part of the 1.0.0 milestone.

I need to keep it open to not miss the things that remain to be done.

PR #281, opened 2020-10-21
Introduce VersionsPropertiesModel

### LINKED ISSUE #256: Keys should be sorted in versions.properties
**Is your feature request related to a problem? Please describe.**

In https://github.com/LouisCAD/kotlin-libraries-playground we have lots of people bringing pull requests, usually adding a dependency in `build.gradle.kts` AND a key in `versions.properties`

If I have to merge two pull requests, the second will systematically have a merge conflict

image

**Using the samples**

Just have a look at the PRs at https://github.com/LouisCAD/kotlin-libraries-playground/pulls
 
We should have a VersionPropertiesWriter or something that would make sure everything is written in a stable sorting order

LouisCAD:
Once ready, will fix #256 

Will also allow to put custom comments, and metdata comments, paving the way to complete the checklist in #134.

Lastly, it will hold the version that generated the file, useful for auto-migration.


jmfayard:
I don't really like how exceptions are used in this current implementation @LouisCAD 
Exceptions are for exceptional, catastrophical errors from which it doesn't make sense to recover. 
That should almost never be the case for parsing a .properties file.
Let's not break the user's build just because she deleted comment lines or something trivial like that. 

I invite you to read this gudance from Roman Elizarov
> You should design your own general-purpose Kotlin APIs in the same way: use exceptions for logic errors, type-safe results for everything else. Don’t use exceptions as a work-around to sneak a result value out of a function.

https://medium.com/@elizarov/kotlin-and-exceptions-8062f589d07

LouisCAD:
I already read the guidance from Roman Elizarov, and I agree with it.

Now, I think it's fair to prevent the users, from putting bad input in a structured file, especially since our users are developers that can just fix the problem themselves quite easily.

To me, exceptions are the simplest way to protect our parser from behaving unpredictably because of incorrect format.

If you have a good example on how to design it without exceptions without making the code dramatically more complex, please show me or tell me.

About being lenient, I disagree. Kotlin file format is not lenient, and that's one of the reasons many people like it. Either it's correct, or in comments that are not breaking expressions/declarations/statements, or it's incorrect and it fails to run, pinpointing the issue.

BTW, the space (` `) is considered as a key value separator in the raw java properties format, just like `:` and `=`, but we're not raw java properties format since we have metadata comments and specific ordering and writing. We just use this format to have syntactic coloration highlight differently the version key, the version value and the comments.

jmfayard:
Kotlin is a turing-complete programming language, it's a different beast that a java properties file, where each line can be understood independantly of the others. Gradle just ignores bad lines and I think we should do the same.

Throwing an exception here is like throwing in `DateFormat.parse()` in the article from Roman.

Instead I would just add a `BadInput` class

```
sealed class Section {
    class BadInput(input: String) : Section()
}
```

LouisCAD:
> Kotlin is a turing-complete programming language
Turing-complete is not really relevant here IMHO, but let's put Kotlin aside if you wish.

> it's a different beast that a java properties file, where each line can be understood independantly of the others.
As you know, that's not the case at all for the convention we have for `versions.properties`. The order of the lines matter.
As of version 0.9.7, we don't allow any user edit apart from changing the used version. The file is systematically overwritten.

In this regard, this PR brings a lot more flexibility by allowing to put comments everywhere it doesn't get in the way of the proper operation.

> Gradle just ignores bad lines and I think we should do the same.

Well, because any plugin might pick-up a property in the `gradle.properties` file, so it cannot really know if you made a mistake. It's almost impossible to make an incorrect java properties file because after it finds a key/value separator character, it treats the rest of the line as the value of the key before.

Let me remind you that we're not making a vanilla java properties file, we're kinda abusing it so we don't have to define our own file format and make an IDE plugin to get proper highlighting. We're using conventions that go beyond the java properties format, and I think that's fine because the `versions.properties` file is only meant to be edited by our plugin and our users.

And since it's only meant to be interpreted by our plugin, I think it's really important to catch mistakes early, I don't see any downside in this particular case, and I don't buy into not breaking Gradle sync because I think it's very easy to not mess up the `versions.properties` file expected format. In other words, I believe that errors in the `versions.properties` ordering are expectional and not something our users which are developers on a dev machine will ever encounter unless they try to test the limits, hence me treating it with exceptions in a rather simple way.

Finally, one could argue that we might use warnings instead of errors, but warnings never prevented NPEs from making it to production, and I know a lot of people have a tendency to ignore these for many reasons, while errors require you to look at them and not let that problem be. I'd approach the issue differently if we were making a library for use outside of developer tools where the problem can be fixed right away BTW.

LouisCAD:
I've thought about your proposal of adding a `BadInput` entry to the `Section` sealed class and I don't like for these 2 reasons:
- it lets carry obvious mistakes that could be fixed right away by the developer
- it doesn't help when there's bad input between parts of a version entry (e.g. bad input between the version key/value pair and an available version comment)

jmfayard:
Fine, let's have it in its current form and we can revisit the issue if stacktraces happens too often.

LouisCAD:
Yes, I'm all in to revisit it if users are annoyed by this or if it proves error-prone with a not great UX.

What do you think about the tests I added?

jmfayard:
I look at it tomorrow

jmfayard:
The new VersionsPropertiesModel is not yet used right?
I wanted to try it on the playground, but I still got the old file header.

jmfayard:
I had a look

It's good to have tests !

There is a test for version = blank in `version-key-missing-value.properties` it would be nice to have one for `version=_`

I have nitpicking about the size of the unit tests (almost 200 lines), they could be much smaller IMHO. Will add comments

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/versions/VersionsPropertiesModel.kt
```diff
@@ -0,0 +1,79 @@
...
+            }
+        }
+
+        data class VersionEntry(
+            val leadingCommentLines: List<String>,
+            val key: String,
+            val currentVersion: String,
+            val availableUpdates: List<String>,
+            val trailingCommentLines: List<String>
+        ) : Section() {
```
jmfayard, line 27: For unit tests, my philosophy is that the smaller there are, the better. 
Instead of having `private fun VersionsPropertiesModel.Section.toText(stringBuilder)`,
I would simply override the `toString()` method for every subclass of section.
This way a test can be written simply as 

```kotlin
val entry = VersionEntry(key = "version.kotlin", currentVersion = "1.4")
"$entry" shouldBe "version.kotlin=1.4"
```

plugins/core/src/test/kotlin/de/fayard/refreshVersions/core/versions/VersionsPropertiesModelTest.kt
```diff
@@ -0,0 +1,83 @@
...
+
+    @Test
+    fun `test parsing old format`() {
+        samplesDir.resolve("old-format").listFiles()!!.filter {
+            it.isDirectory
+        }.also { check(it.isNotEmpty()) }.forEach {
+            val inputFile = it.resolve("input.properties")
+            val outputFile = it.resolve("output.properties")
+            val parsedModel = VersionsPropertiesModel.readFromText(inputFile.readText())
+            assertEquals(
```
jmfayard, line 24: Why not using Kotest matchers?
They are very nice IMHO https://kotest.io/matchers/core/

jmfayard (approved):

jmfayard:
How does the parser handle it when multiple sections have the same key?

LouisCAD:
> The new VersionsPropertiesModel is not yet used right?
> I wanted to try it on the playground, but I still got the old file header.

Yes, it's not used yet. I plan to plug it this weekend.

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/versions/VersionsPropertiesModel.kt
LouisCAD: `toString()` is more targeted at debug printing when it's not about `CharSequence` or numbers. I'm afraid it would be confusing to have the `toString()` be part of the "business" logic, especially in a `data class`.

I don't disagree on making tests simple, but I don't think smaller is always simpler.

I'll see if the tests can be made simpler though.

LouisCAD:
> I had a look
> 
> It's good to have tests !
> 
> There is a test for version = blank in `version-key-missing-value.properties` it would be nice to have one for `version=_`
> 
> I have nitpicking about the size of the unit tests (almost 200 lines), they could be much smaller IMHO. Will add comments

You mean the size of the test files data?

I've been thinking about this as well.

I'll consider shortening a bunch of them.

LouisCAD:
> How does the parser handle it when multiple sections have the same key?

Right now, the parser doesn't check for duplicate keys at all, but that's something I'll need to take into account.

As for how it's currently behaving in refreshVersions for duplicate keys in `versions.properties`, it takes the one `java.util.Properties` considered.

I think I'd ensure refreshVersions doesn't bring duplicate version keys (that was already planned), and for duplicate keys, I think I'll outlaw this, which is fine IMHO since the IDE already shows an error on duplicate keys in `.properties` files. That way, the users will have to delete one of the duplicate entries, and will select/merge/edit any surrounding comments themselves, without us having to take a potentially wrong decision regarding comments preservation/merging.

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/versions/VersionsPropertiesModel.kt
jmfayard: I'm fine with `toString()` but if not just declare an abstract method `toText() : CharSequence` on the `Section` class and the same thing works

plugins/core/src/test/kotlin/de/fayard/refreshVersions/core/versions/VersionsPropertiesModelTest.kt
LouisCAD: I was not familiar with them, that's why.

Now, I found them pretty straightforward, which is why you'll find them in my latest push.

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/versions/VersionsPropertiesModel.kt
LouisCAD: Regarding the current `toText()` function, why do you think it'd be better to have it as an abstract function in the `Section` sealed class instead of an extension as it currently is?

I personally chose the extension function approach to not overcrowd the model file and limit it to arguments checking and related constants, putting the read and write logic externally.

jmfayard: It's fine to have one single `Section.toText()` method.

plugins/core/src/test/kotlin/de/fayard/refreshVersions/core/versions/VersionsPropertiesModelTest.kt
LouisCAD: FYI, I reverted this commit because the output of kotest is different from junit5 backed kotlin.test, and as a result, failures are not parser by the IDE where it'd give you the nice diff view.

LouisCAD:
> The new VersionsPropertiesModel is not yet used right?
> I wanted to try it on the playground, but I still got the old file header.

Now it's used!

First improvement, apart from keeping comments: adding a dependency managed by refreshVersions puts the new key in the alphabetical order straightaway.

jmfayard:
I tried the plugin in the Libraries Playground and found this bug

https://github.com/LouisCAD/kotlin-libraries-playground/commit/cb134016e4ed0e0012d1ded17dfaf3916c52839c

```
 Suppress version keys and check whether they are recreated

1. Suppress some version keys
2. Sync Gradle in IntelliJ

It fails with this error
`java.lang.IllegalStateException: @NotNull method org/jetbrains/plugins/gradle/tooling/ErrorMessageBuilder.getRootCauseMessage must not return null`

See https://gradle.com/s/5ww6jkvi5b42m

 $ ./gradlew --scan hello
doesn't crash https://gradle.com/s/5voaceh3hqfl4

but it doesn't recreate the keys either
```

jmfayard:
I suppressed the header comments to see what happened and got a stracktrace

https://github.com/LouisCAD/kotlin-libraries-playground/commit/b83e7477f6fb4e2af5fbe7802d57e14672cd7477

```
 Suppress the comments header and run $ ./gradlew refreshVersions

Execution failed for task ':refreshVersions'.
> Didn't find the passed prefix into the given String

https://gradle.com/s/2qmivmpsatsds
```

I don't like the idea to rely on comments parsing to find the version, it feels like it should be a key (open for the naming)

```
# versions.properties
generated.with.refreshVersions=0.9.8-SNAPSHOT
``

**Comment by jmfayard** [comment]
At https://github.com/LouisCAD/kotlin-libraries-playground/commit/c7565b19e2128e3a11940f890237b5862bb7b791 I have run successfully `$ ./gradlew refreshVersions` 

I notice that like before it generates `version.ktor` with a leading space

```properties
version.kotlinx.serialization=1.0.0

 version.ktor=1.4.1

version.moshi=1.11.0
```

I think the version properties model should trim the version key and value

jmfayard:
At commit https://github.com/LouisCAD/kotlin-libraries-playground/commit/29796a7ecff4d5d77a1c0203499629ce007fe6f4 I tried the behavior when you have a duplicate version key

```properties
version.moshi=1.10.0

version.moshi=1.11.0
```

The current behavior is this :
- `$ ./gradlew refreshVersions` shows a stacktrace  `duplicate key version.moshi`
- *but* a simple Gradle sync will silently works, and in my example at least take the highest version

image

As a user, I don't like this behavior. I would prefer that the plugin takes care of resolving the situation with a single version, the highest of both. It's almost always what I want, and in the rare cases where it's not, I will see it in my `$ git diff`

jmfayard (commented):
Done with comments

LouisCAD:
> At [LouisCAD/kotlin-libraries-playground@c7565b1](https://github.com/LouisCAD/kotlin-libraries-playground/commit/c7565b19e2128e3a11940f890237b5862bb7b791) I have run successfully `$ ./gradlew refreshVersions`
> 
> I notice that like before it generates `version.ktor` with a leading space
> 
> ```ini
> version.kotlinx.serialization=1.0.0
> 
>  version.ktor=1.4.1
> 
> version.moshi=1.11.0
> ```
> 
> I think the version properties model should trim the version key and value

The leading space is for short keys, to keep the alignment when there's the available comments. I kept this behavior to avoid git diff issues where it thinks the whole line changed because the alignment had to change to match the available version comments.

LouisCAD:
> At commit [LouisCAD/kotlin-libraries-playground@29796a7](https://github.com/LouisCAD/kotlin-libraries-playground/commit/29796a7ecff4d5d77a1c0203499629ce007fe6f4) I tried the behavior when you have a duplicate version key
> 
> ```ini
> version.moshi=1.10.0
> 
> version.moshi=1.11.0
> ```
> 
> The current behavior is this :
> 
> * `$ ./gradlew refreshVersions` shows a stacktrace  `duplicate key version.moshi`
> * _but_ a simple Gradle sync will silently works, and in my example at least take the highest version
> 
> image
> 
> As a user, I don't like this behavior. I would prefer that the plugin takes care of resolving the situation with a single version, the highest of both. It's almost always what I want, and in the rare cases where it's not, I will see it in my `$ git diff`

The highest version might not always be the right choice if you're doing Kotlin Multiplatform with kotlinx.coroutines for example. For that reason, I think it's best to fail and tell the user why.

Now, the reason why it doesn't fail at Gradle sync is because there's no parser (it's currently only used when there's a need to write the file). I think I should change that to have it run everytime the file is read so we catch errors early.

jmfayard:
That something doesn't work 100% of the time should not prevent us for doing the better thing most of the time. I would let `git diff` handle the corner cases

jmfayard:
It's strange that the parser is not used during the sync, but that there is a regression when you suppress the version keys https://github.com/jmfayard/refreshVersions/pull/281#issuecomment-716391604

LouisCAD:
> I suppressed the header comments to see what happened and got a stracktrace
> 
> [LouisCAD/kotlin-libraries-playground@b83e747](https://github.com/LouisCAD/kotlin-libraries-playground/commit/b83e7477f6fb4e2af5fbe7802d57e14672cd7477)
> 
> ```
>  Suppress the comments header and run $ ./gradlew refreshVersions
> 
> Execution failed for task ':refreshVersions'.
> > Didn't find the passed prefix into the given String
> 
> https://gradle.com/s/2qmivmpsatsds
> ```
> 
> I don't like the idea to rely on comments parsing to find the version, it feels like it should be a key (open for the naming)
> 
> ```ini
> # versions.properties
> generated.with.refreshVersions=0.9.8-SNAPSHOT
> ```

I need to handle that error with a proper error message.

Now, it's explictly written to not remove the header comments, so I have no problem failing if the developers proceed anyway.

I used a comment because it's less constraining than the java properties keys where you cannot use spaces and a bunch of other characters.

Do you think our users would remove this header comments against that warning, expect the thing to keep working, and complain about that?

jmfayard:
> Do you think our users would remove this header comments against that warning, expect the thing to keep working, and complain about that?

I would reframe it this way :
I never expect a software to behave differently when you add or remove comments. 
I remember Google doing such a hack in the early days of JavaScript where they would put the static type of the parameters in a commment. 
Now you have TypeScript where the types are part of the source code.
That makes much more sense IMHO.

LouisCAD:
> It's strange that the parser is not used during the sync, but that there is a regression when you suppress the version keys [#281 (comment)](https://github.com/jmfayard/refreshVersions/pull/281#issuecomment-716391604)

Yep, I saw your comment, I will investigate.

LouisCAD:
> > Do you think our users would remove this header comments against that warning, expect the thing to keep working, and complain about that?
> 
> I would reframe it this way :
> I never expect a software to behave differently when you add or remove comments.
> I remember Google doing such a hack in the early days of JavaScript where they would put the static type of the parameters in a commment.
> Now you have TypeScript where the types are part of the source code.
> That makes much more sense IMHO.

And Typescript came with its own file format, and tons of investments followed to make it supported widely.

At least, in our case, we have a parser that can check the file correctness (maybe Google did it as well before TS?).

Now, do you think we should abandon the java properties format and use something else that better suits our needs and is already supported by IDEs? If so, I'll be happy to evaluate the available solutions.

jmfayard:
Nah, `versions.properties` is fine! The time to change will come only when Gradle comes up with its own format. 

Also once the plugin of settings is done, we can use an extension `refreshVersions { ... }` to configure the plugin

LouisCAD:
To summarize our discussion above, I have to:
1. Investigate the issue where removed version entries don't come back at Gradle sync.
2. Make a better error message for when the header is removed or tampered with by the developer.
3. Run the parser on every build instead of solely using `java.util.Properties`, so errors surface early.

I think the other matters are complicated and might be better discussed over the phone to settle the decisions.

Is there anything else I'd have to do so far?

LouisCAD:
@jmfayard I just pushed small a change to ensure the `versions.properties` file is checked on every read, which means on every Gradle reload/sync/import or build.

Now, I'm unable to reproduce the issue where removed entries don't come back at Gradle reload/sync/import… do you have reproducing steps so I can trigger the issue and debug it?
Or does the bug no longer reproduce?

jmfayard:
@LouisCAD bug fixed, I have checked !

LouisCAD:
That's really weird because I don't think I changed much or anything that should relate to that…

I will synchronize with a shared lock all accesses to the `versions.properties` file because I suspect not going so might cause race conditions when Gradle runs in parallel mode, but I really don't think that's the issue we saw earlier.

Issue #276, opened 2020-10-18
Have a staging website


LouisCAD:
To review website updates more comfortably (i.e. not developing directly in production), we need to be able to deploy to another domain.

GitHub pages don't seem to make this easy, so I'm considering using Firebase Hosting for that (its free plan will be more than enough for PR reviews).

I found this which should easily be usable in GitHub Actions (despite it showing an example for Gitlab): https://github.com/tkikuchi2000/mkdocs-firebase-pipeline

jmfayard:
I gave it some thoughts.

I think your real issue can be solved in the future by:

1. not pushing directly to `master`/`mkdocs` 
2. running `$ mkdocs serve` locally when reviewing a pull request 

When the website was in its infancy, it made sense to push directly to production, because higher velocity and no-one was using it anyway.

Once we it has replaced the current docs and people use it, running mkdocs serve locally is a superior solution. A staging website would be read by no-one except the pull request reviewer. By running the website locally you get that too. But much important you can update the docs that need to be updated and see in love whether that looks good.

LouisCAD:
I think both options are interesting. The ability to not have to check out a PR to see edits is nice to me, and more convenient to also preview on mobile.

LouisCAD:
I added a "workflow" in #280 so it now uploads the built website as artifacts that can be downloaded as a zip file.
Then, using the website locally doesn't require to install mkdocs and python stuff on your machine, all I have to do is:
1. Unzip the downloaded artifact
2. Drag and drop the unzipped directory to IntelliJ IDEA
3. Open the `index.html` file in the IDE
4. Click on the icon of my favorite browser to launch the IntelliJ IDEA basic web server and have it resolve the index.html pages.

LouisCAD:
Regarding Firebase Hosting, I've been exploring it a little more and here's what I found that is interesting for us:
1. Even though Firebase as a Google product is not really available in China (unlike GitHub), websites using Firebase Hosting and a custom domain are reachable through the firewall
2. We can get server-side and CDN-side analytics with Firebase Hosting
3. There's [some docs about deploying via GitHub Actions](https://firebase.google.com/docs/hosting/github-integration) for preview and/or live channels, and [a GitHub Action](https://github.com/marketplace/actions/deploy-to-firebase-hosting) that we can leverage.
4. Firebase Hosting has a free tier allowing custom domains that should be enough for us (I guess you knew).

jmfayard:
We can use ngrok for that

PR #284, opened 2020-10-28
add google cloud storage support

### LINKED ISSUE #283: Google cloud storage based maven repositories
**Is your feature request related to a problem? Please describe.**

adding repositories hosted on google cloud storage like this is rather popular in organizations

```gradle-kts
maven("gcs://mvn-bucket/releases") {
	name = "$organization maven"
}
```

as part of my next task of cleaning up buildscripts
to migrate the work backend and android app to refreshVersions i need this and wo

**Additional context**

not sure how gradle does it but..
you need to have `gcloud` installed on the system
and the env var `GOOGLE_APPLICATION_CREDENTIALS` pointing to a json file containing the service account
see: https://cloud.google.com/docs/authentication/getting-started

seems like google directly provides a sdk: `com.google.cloud:google-cloud-storage:_`
and it seems easy to use, this seems to be the only code necessary to read and write files from gcs (but we only need to read i guess)
https://github.com/tehlers/gradle-gcs-build-cache/blob/master/src/main/kotlin/net/idlestate/gradle/caching/GCSBuildCacheService.kt

NikkyAI:
fixes #283 

~~rough first attempt, but it works~~

refactored and also added support for `mavenLocal()` and other `file://` based repos


plugins/core/build.gradle.kts
```diff
@@ -38,6 +38,7 @@ dependencies {
         because("It has ready to use HttpException class")
     }
     implementation(Square.moshi.kotlinReflect)
+    implementation("com.google.cloud:google-cloud-storage:1.113.2")
```
LouisCAD, line 41:
```suggestion
    implementation("com.google.cloud:google-cloud-storage:_")
```
I think you expected that first feedback!

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/MavenDependencyVersionsFetcherGCS.kt
```diff
@@ -0,0 +1,61 @@
...
+            lastUpdateTimestampMillis = parseLastUpdatedFromMavenMetaData(xml),
+            availableVersions = if (versionFilter == null) {
+                allVersions
+            } else {
+                allVersions.filter(versionFilter)
+            }
+        )
+    }
+
+    companion object {
```
jmfayard, line 43: Isn't that code a duplicate of the standard MavenDependencyVersionsFetcher?

NikkyAI: it is, and thats why i would call it a hack until i refactor it
i image standard MavenDependencyVersionsFetcher can just be implemented in 3 ways
1. httpCLient
2. reading files (for mavenLocal)
3. using gcn

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/MavenDependencyVersionsFetcherGCS.kt
```diff
@@ -1,61 +1,83 @@
...
-                .split("<version>", "</version>")
-                .mapNotNull { if (it.isBlank()) null else Version(it.trim()) }
+            throw FileNotFoundException("Unable to load '$fullPath' from Google Cloud Storage bucket '$bucketName'.")
         }
+
+    }
+
+    companion object {
+        private val CREDENTIALS_PATH = System.getenv("GOOGLE_APPLICATION_CREDENTIALS")
+            ?: error("required environment variable 'GOOGLE_APPLICATION_CREDENTIALS' missing")
```
LouisCAD, line 81: Is that a standard or something specific to your project's or company's conventions?
If it's standard, please comment with the link here.

LouisCAD (changes requested):
@NikkyAI There's compilations error right now. I need them to be fixed in order to review the changes.

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/MavenDependencyVersionsFetcherGCS.kt
NikkyAI: https://cloud.google.com/docs/authentication/getting-started#setting_the_environment_variable

NikkyAI: it is the same as what gradle expects there to be

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/MavenDependencyVersionsFetcherGCS.kt
NikkyAI: now i refactored it so that there should not be duplicate

jmfayard:
@NikkyAI 
Can you copy the maven repository from Ben Manes's versions plugin ? 
This way we can have unit tests for the `file://` part !
https://github.com/ben-manes/gradle-versions-plugin/tree/master/src/test/resources

Apart from that, I won't be able to help much, I'm moving to Paris this week-end

plugins/versions.properties
```diff
@@ -30,3 +30,5 @@ version.okhttp3=4.8.1
 version.retrofit2=2.9.0
 
 version.moshi=1.9.3
+
+version.com.google.cloud..google-cloud-storage=1.113.2
```
jmfayard, line 34: Run `$ ./gradlew refreshVersions` inside `plugins`, this way the file will be correctly sorted and you won't have a merge conflict

jmfayard (approved):

LouisCAD (changes requested):
This PR breaks existing functionality: non existing files in a file repo (e.g. `mavenLocal`) would get an `HttpException` to be thrown with the 404 code via the OkHttp interceptor, but this change changes the types of the exceptions to be the raw ones (`FileNotFoundException`). Consequently, I'll need to perform other changes or reachitect the changes.

**EDIT:** I saw the code was catching `FileNotFoundException` so it wasn't breaking it, but it was duplicating functionality.

LouisCAD:
Also using `@ExperimentalStdlibApi` is fine in end user apps, but it's not in plugins and libraries that need to keep binary compatibility whenever possible (like `refreshVersions`).

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/MavenDependencyVersionsFetcherGCS.kt
```diff
@@ -0,0 +1,71 @@
...
+                .service
+
+            bucket = storage.get(bucketName) ?: throw BuildCacheException("$bucketName is unavailable")
+
+        } catch (e: FileNotFoundException) {
+            throw GradleException("Unable to load credentials from $CREDENTIALS_PATH.", e)
+        } catch (e: IOException) {
+            throw GradleException("Unable to access Google Cloud Storage bucket '$bucketName'.", e)
+        } catch (e: StorageException) {
+            throw GradleException("Unable to access Google Cloud Storage bucket '$bucketName'.", e)
```
LouisCAD, line 45: `GradleException` is not designed for exceptions from plugins but for exceptions from Gradle itself.

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/MavenDependencyVersionsFetcherGCS.kt
```diff
@@ -0,0 +1,71 @@
...
+                .setCredentials(ServiceAccountCredentials.fromStream(FileInputStream(CREDENTIALS_PATH)))
+                .build()
+                .service
+
+            bucket = storage.get(bucketName) ?: throw BuildCacheException("$bucketName is unavailable")
+
+        } catch (e: FileNotFoundException) {
+            throw GradleException("Unable to load credentials from $CREDENTIALS_PATH.", e)
+        } catch (e: IOException) {
+            throw GradleException("Unable to access Google Cloud Storage bucket '$bucketName'.", e)
```
LouisCAD, line 43: The exceptions caught here are obscure. Why would a `FileNotFoundException` be thrown and by which method? There's no mention about this in the methods appearing in the `try` block.

Also, `GradleException` is not designed for exceptions from plugins but for exceptions from Gradle itself.

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/MavenDependencyVersionsFetcherGCS.kt
```diff
@@ -0,0 +1,71 @@
...
+        } catch (e: StorageException) {
+            throw GradleException("Unable to access Google Cloud Storage bucket '$bucketName'.", e)
+        }
+    }
+
+    @OptIn(ExperimentalStdlibApi::class)
+    fun getBlob(path: String): String {
+        val fullPath = listOfNotNull(repoPath, path).joinToString("/", "", "")
```
LouisCAD, line 51: The last two parameters are not needed because they are already the default values.
If they had to be defined, it'd be way better to use named arguments instead of throwing same typed arguments in there without any indication about what is what.

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/MavenDependencyVersionsFetcherGCS.kt
```diff
@@ -0,0 +1,71 @@
...
+        try {
+            val blob = bucket.get(fullPath) ?: throw FileNotFoundException("blob $fullPath is missing")
+            return blob.getContent().decodeToString()
+
+        } catch (e: StorageException) {
+            // https://github.com/googleapis/google-cloud-java/issues/3402
+            if (e.message?.contains("404") == true) {
```
LouisCAD, line 58: Brittle. Can be replaced by this (which I'll do):
```kotlin
val cause = e.cause
if (cause is GoogleJsonResponseException && cause.statusCode == 404)
```

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/MavenDependencyVersionsFetcherGCS.kt
```diff
@@ -0,0 +1,71 @@
...
+                throw FileNotFoundException("missing blob")
+            }
+
+            throw FileNotFoundException("Unable to load '$fullPath' from Google Cloud Storage bucket '$bucketName'.")
```
LouisCAD, line 62: I don't think it's the right exception type for this case.

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/MavenDependencyVersionsFetcherGCS.kt
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+    }
+
+    companion object {
+        private val CREDENTIALS_PATH = System.getenv("GOOGLE_APPLICATION_CREDENTIALS")
```
LouisCAD, line 68: This means it'd fail for users that use public to read dependencies hosted on Google Cloud Storage… did you consider this?

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/MavenDependencyVersionsFetcherFILE.kt
```diff
@@ -0,0 +1,22 @@
...
+    moduleId: ModuleId,
+    repoUrl: String
+) : MavenDependencyVersionsFetcher(
+    moduleId = moduleId,
+    repoUrl = repoUrl
+) {
+    private val baseFolder = File(repoUrl)
+
+    init {
+//        require(repoUrl.endsWith('/'))
```
LouisCAD, line 16: There's no reason to leave commented out code. It's confusing to have in a codebase.

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/DependencyVersionsFetcher.Companion.kt
```diff
@@ -15,32 +15,31 @@ internal operator fun DependencyVersionsFetcher.Companion.invoke(
...
-        repoAuthorization = when (repository.name) {
-            "MavenLocal" -> null // MavenLocal never needs authorization.
-            // Getting credentials on mavenLocal results in insanely
-            // hard to  diagnose error that happens later with an unrelated stacktrace.
-            // See this issue: https://github.com/jmfayard/refreshVersions/issues/222
-            // The error message would be the following:
-            // "Authentication scheme 'all'(Authentication) is not supported by protocol 'file'"
-            else -> repository.passwordCredentials?.let { credentials ->
+    return when (repository.url.scheme) {
+        "https" -> MavenDependencyVersionsFetcherHTTP(
```
LouisCAD, line 19: This naming doesn't follow Kotlin official style guide which we're using.

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/DependencyVersionsFetcher.Companion.kt
```diff
@@ -15,32 +15,31 @@ internal operator fun DependencyVersionsFetcher.Companion.invoke(
...
                 )
             }
-        }
-    )
+        )
+        "file" -> MavenDependencyVersionsFetcherFILE(
+            moduleId = ModuleId(group, name),
+            repoUrl = repository.url.toString()
+        )
+        "gcs" -> MavenDependencyVersionsFetcherGCS(
```
LouisCAD, line 34: Doesn't follow Kotlin official naming conventions either.
We'll use `MavenDependencyVersionsFetcherGoogleCloudStorage` to avoid leaving domain specific acronyms in the codebase that might confuse future contributors for a bit.

LouisCAD (changes requested):
Please, answer the questions I asked, but don't push to the branch, I'll finish the work (already started changes).

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/MavenDependencyVersionsFetcherGCS.kt
LouisCAD: Also, there's no need to use `listOfNotNull` because the arguments are not nullable… and using `joinToString` is overkill and obscures the code as there's only two `String`s to join, which means a string template would work as well, even a little more efficiently actually.

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/MavenDependencyVersionsFetcherGCS.kt
LouisCAD: FYI, `StorageException` is annotated with `@InternalApi` (in Java, so no warning…). I'm looking for alternatives.

LouisCAD:
My changes on top of yours so far can be compared with the current `develop` branch here: https://github.com/jmfayard/refreshVersions/compare/develop...gcs-support-pr-fixes?expand=1

Please @NikkyAI, test it and let me know if the GCS integration works properly, and if there's any gotchas that need to be taken into account or other things that can be better handled.

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/MavenDependencyVersionsFetcherGCS.kt
NikkyAI: i think there is no way to resolve a bucket name without a service account or credentials binding yourself to a user ?

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/MavenDependencyVersionsFetcherGCS.kt
NikkyAI: i think that is the one that is being caught by refreshVersions, other exceptions will crash when a depedency is not available on gcs

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/MavenDependencyVersionsFetcherGCS.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    init {
+        try {
+
+            val storage = StorageOptions.newBuilder()
+                .setCredentials(ServiceAccountCredentials.fromStream(FileInputStream(CREDENTIALS_PATH)))
+                .build()
+                .service
+
+            bucket = storage.get(bucketName) ?: throw BuildCacheException("$bucketName is unavailable")
```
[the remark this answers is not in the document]
NikkyAI: oops, i copied fro mthe gcs buildcache plugin, seems like that slipped through

NikkyAI:
i tested it and i am running into errors like 
```
java.lang.IllegalStateException: java.lang.IllegalArgumentException: Expected a comment but found random text:
```

i completely removed the versions.properties file and let it regenerate as much as it could but that still lead to the same content (with a different comment at the top)

sometimes i am getting `Putting custom comments between available updates comments is not supported.`

LouisCAD:
@NikkyAI I'll need a reproducing `versions.properties` for troubleshooting and improvements.

NikkyAI:
i can verify that with a few fixes for windows specifc issues it works well

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/MavenDependencyVersionsFetcherGCS.kt
LouisCAD: I did a refactor so there's hidden contract about which exceptions mean not found.

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/MavenDependencyVersionsFetcherGCS.kt
LouisCAD: We can use `NoCredentials.getInstance()`. Did the edit.

LouisCAD (approved):
Did a review with my changes included. Looks good to me.

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/MavenDependencyVersionsFetcherFile.kt
```diff
@@ -0,0 +1,35 @@
...
+) {
+    private val repoDir = File(repoUrl.substringAfter("file:"))
+
+    init {
+        require(repoUrl.endsWith('/'))
+    }
+
+    override suspend fun getXmlMetadataOrNull(): String? = try {
+        val targetDir = repoDir.resolve("${moduleId.group!!.replace('.', '/')}/${moduleId.name}")
+        val fileNames: Array<String>? = targetDir.list()
```
jmfayard, line 22: I suggest a quick return here to avoid using `?.` everywhere
`targetDir.list() ?: return null`

jmfayard (approved):

jmfayard:
Is there a public google cloud URL so that we can test that it works ?

NikkyAI:
i can verify that the code in its current state works, i can set up a storage bucket and create a readonly service account later on my account

LouisCAD:
@jmfayard Maybe we can make a joint Firebase/GCP project (which brings access to Google Cloud Storage) and put the scoped service accounts secrets in GitHub secrets to power GCS testing in GitHub Actions + on our machines.
We would then use this Firebase project for the website as well (see #276).

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/MavenDependencyVersionsFetcherFile.kt
NikkyAI: applied that suggestion

LouisCAD (approved):

NikkyAI:
so, can this be merged and added to a release if there is no more breaking things ?

jmfayard:
I'm sorry Nikky, I know it's frustrating, we need to improve our release process. Hopefully we can have a development version released soon

LouisCAD:
@jmfayard Maybe we can put the GCS support behind a flag (e.g. `refreshVersions.preview.gcs`), so we get a dev release for Nikky to use, and us to benefit from the other great changes of this PR, like the fix for mavenLocal repos?

jmfayard:
I'm fine with that

NikkyAI:
well gcs is not esen used without a gcs repo being configured.. so i would just print a warning when it is used for now

jmfayard:
@NikkyAI @LouisCAD  we can now use jitpack to have dev versions published faster

See #290 https://github.com/jmfayard/refreshVersions/issues/290

For this PR it would be confirmed this way in `settings.gradle.kts` 

```kotlin
buildscript {
    val githubPR: Int? = 284
    repositories {
        gradlePluginPortal()
        if (githubPR != null) maven("https://jitpack.io")
    }
    val classpath = if (githubPR == null)
        "de.fayard.refreshVersions:refreshVersions:0.9.7" else
        "com.github.jmfayard:refreshVersions:PR${githubPR}-SNAPSHOT"
    dependencies.classpath(classpath)
}
```

NikkyAI:
this dependency on jitpack seems to work for me: `com.github.jmfayard:refreshVersions:feature~gcs-SNAPSHOT`but then https://github.com/jmfayard/refreshVersions/blob/9d862c71c4cf871d10770a836a25600f3737d68a/plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/CurrentlyUsedVersion.kt#L8 does not match

LouisCAD:
@NikkyAI Why did you force-push a change instead of simply pushing a plain commit?

NikkyAI:
i force pushed undid my changes

NikkyAI:
it currently seems to break android plugin because of dependencies
can we shadow and relocate gcs and dependencies ?

LouisCAD:
@NikkyAI Shadowing dependencies has big implications when it comes to plugin size, and we want to avoid delaying ephemeral CIs (such as GitHub Actions) further for projects using refreshVersions for a feature that many projects will not need.

There might be another way to avoid dependency version conflicts, we'll need to test it.

FYI, I'll get back to this early next year. We will test something with another optional module, and if it works, we might use that same technique for this feature.

LouisCAD:
Hello, just to let you know I added this PR into my priority list (currently number 3).

I think we can have an extra module that brings GCS support, and a clever way to enabled it very easily, in a discoverable manner. I'll experiment with that soon.

LouisCAD:
> it currently seems to break android plugin because of dependencies
@NikkyAI Please, can you give more details on the actual breakage, and the AGP version(s) you witnessed the issues with?

LouisCAD:
Thanks for your help on Kotlin's Slack in reproducing this and finding a solution.

I'm going to push the changes I made to avoid the issue from triggering.

LouisCAD:
Again, thank you very much @NikkyAI for your contribution, your patience and your help!

I think GCS support is very good addition, especially considering the fact that Google is the most environment conscious and friendly companies of its scale in the cloud computing area, and it's been so for a long time (close to 2 decades back).

I think GCS support in refreshVersions has the potential to be very helpful for teams, developers and companies that have private artifacts to store.

PR #287, opened 2020-11-01
change version.txt from symlink to generated-resource folder

NikkyAI:
this should make it deployable on windows too


plugins/core/build.gradle.kts
```diff
@@ -62,6 +63,18 @@ kotlin {
     javaComponent.withVariantsFromConfiguration(configurations["testFixturesRuntimeElements"]) { skip() }
 }
 
+val genResources = buildDir.resolve("generated/refreshVersion/resources")
```
LouisCAD, line 66:
```suggestion
val genResources = buildDir.resolve("generated/refreshVersions/resources")
```

plugins/core/build.gradle.kts
```diff
@@ -62,6 +63,18 @@ kotlin {
...
+
+sourceSets {
+    main {
+        resources.srcDir(genResources.path)
+    }
+}
```
LouisCAD, line 72:
```suggestion
sourceSets.main {
    resources.srcDir(genResources.path)
}
```

LouisCAD:
Thank you! Made a few edits, merging.

PR #291, opened 2020-11-11
WIP: allows using plugins without marker

NikkyAI:
this is something i hacked together because i could not figure out how to make use of `pluginManagement.resolutionStrategy.eachPlugin` properly since refreshVersions seems to override it

so instead a lambda is passed to the refreshVersions call

usage:

```
bootstrapRefreshVersions(
    pluginResolution = { pluginId, version ->
        logger.lifecycle("picking group and name for $pluginId with version $version")
        when (pluginId) {
            "com.android.application", "com.android.library" -> "com.android.tools.build:gradle"
            "kotlin-android", "kotlin-kapt", "kotlin-android-extensions" -> "org.jetbrains.kotlin:kotlin-gradle-plugin"
            "androidx.navigation.safeargs.kotlin" -> "androidx.navigation:navigation-safe-args-gradle-plugin"
            "com.google.gms.google-services" -> "com.google.gms:google-services"
            "realm-android" -> "io.realm:realm-gradle-plugin"
            "de.mannodermaus.android-junit5" -> "de.mannodermaus.gradle.plugins:android-junit5"
            else -> null
        }
    }
)
```

### DIFF
--- a/jitpack.yml
+++ b/jitpack.yml
@@ -1,0 +1,15 @@
+# https://jitpack.io/docs/BUILDING/
+# implementation 'com.github.jmfayard:refreshVersions:Tag'
+jdk:
+  - openjdk8
+#before_install:
+#  - ./custom_setup.sh
+install:
+  - echo "implementation(com.github.$GROUP:$ARTIFACT:$VERSION-SNAPSHOT)"
+  - echo "See https://jitpack.io/com/github/$GROUP/$ARTIFACT/$VERSION/"
+  - echo "Logs at https://jitpack.io/com/github/$GROUP/$ARTIFACT/$VERSION/build.log"
+  - cd plugins
+  - ./gradlew publishToMavenLocal
+  - cd ../
+#env:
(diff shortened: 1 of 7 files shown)


LouisCAD:
The `buildscript` way that these plugins without markers require is supported by refreshVersions already FYI.

NikkyAI:
when i tried to use it on 0.97 and 0.98-SNAPSHOT it did not work, maybe verify that with plugins other than the android plugin that has a workaround hardcoded in refreshVersions ?

LouisCAD:
Hello @NikkyAI, is that a feature you still want?

NikkyAI:
it would be nice to have the same syntax of `id("some-plugin")` in the buildscript and define the lookup of the dependency in `pluginManagement.resolutionStrategy`
i have not run into this problem as much recently.. and if i did.. i declared the depenencies in the root project `buildScript.depenencies`

but i think the closer this can model to the usual buildscripts the easier it is to pick up refreshVersions without frustration in complex projects

PR #304, opened 2020-12-11
Revert all dependencies to Kotlin 1.3.72

jmfayard:
https://gradle.com/s/tv3kpbjynyafc
Gradle 6

### DIFF
--- a/plugins/versions.properties
+++ b/plugins/versions.properties
@@ -9,18 +9,55 @@ plugin.com.gradle.plugin-publish=0.12.0
 
 version.junit=5.7.0
 
-version.kotest=4.3.1
+version.kotest=4.1.2
+## # available=4.1.3
+## # available=4.2.0.RC1
+## # available=4.2.0.RC2
+## # available=4.2.0.RC3
+## # available=4.2.0
+## # available=4.2.2
+## # available=4.2.3
+## # available=4.2.4
+## # available=4.2.5
+## # available=4.2.6
+## # available=4.3.0
+## # available=4.3.1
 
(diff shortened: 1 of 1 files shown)


LouisCAD (dismissed):
I guess you did it super quick and bypassed basic local testing… I'll let you find what's wrong.

plugins/versions.properties
```diff
@@ -11,14 +11,16 @@ version.junit=5.7.0
...
+version.kotlinpoet=1.6.0
 
-version.kotlinx.coroutines=1.3.9
+version.kotlinx.coroutines=1.2.1
 
-version.moshi=1.11.0
+version.moshi=1.9.0
 
-version.okhttp3=4.9.0
+version.okio=2.7.0
```
LouisCAD, line 22: Why is this added?

plugins/versions.properties
```diff
@@ -11,14 +11,16 @@ version.junit=5.7.0
...
 version.kotest=4.3.1
 
-version.kotlin=1.4.20
+version.kotlin=1.3.72
 
-version.kotlinpoet=1.7.2
+version.kotlinpoet=1.6.0
 
-version.kotlinx.coroutines=1.3.9
+version.kotlinx.coroutines=1.2.1
```
LouisCAD, line 18: I think we should use version 1.3.8, the latest Kotlin 1.3 compatible release IIRC.

LouisCAD:
You downgraded kotlinx.coroutines more than needed, and kotest was using Kotlin stdlib 1.4.

I fixed it, merging.

PR #308, opened 2020-12-11
github actions: staging website

jmfayard:
https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables


mkdocs.yml
```diff
@@ -64,7 +64,7 @@ nav:
   - 'Setup': setup.md
   - 'Add Dependencies': add-dependencies.md
   - 'Update Dependencies': update-dependencies.md
-#  - 'Gradle buildSrcVersions': gradle-buildsrcversions.md
+  - 'Gradle buildSrcVersions': gradle-buildsrcversions.md
```
LouisCAD, line 67: This change is unrelated to this PR (out of the scope), plus we previously agreed that we would not make it part of the main menu for now.

LouisCAD, line 67: This change is unrelated to this PR (out of the scope), plus we previously agreed that we would not make it part of the main menu for now.

site.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+# USAGE:
+#    site.sh              # serve website locally
+#    site.sh mybranch     # publish to https://jmfayard.github.io/rvdev/
+#    site.sh release      # publish to https://jmfayard.github.io/refreshVersions/
```
LouisCAD, line 5: The word "site" lacks explicitness about what it actually does.

I think this should be split into two scripts:
1. `build-website.sh`
2. `publish-website.sh`

The `publish-website.sh` script would:
1. Require a branch name to be passed as an argument.
2. Run the `build-website.sh` script.
3. Publish it at the right place.

LouisCAD (changes requested):

jmfayard:
the GitHub actions do not even run for unknown reasons so I gave up on this pull request

jmfayard:
marked as draft

PR #315, opened 2021-01-22
Fixing ordered list in the index page

cortinico:
The bullet points on the very first page are off, fixing them:

Screenshot 2021-01-22 at 21 49 36


LouisCAD:
Thank you!

Not deployed just yet, but I'll see if I can cherry-pick it into the release branch so it doesn't wait for next release.

LouisCAD:
Cherry-picked successfully along with what had to be as well, and we can see the fix on the website.
Thanks!

PR #335, opened 2021-05-02
Make buildSrcLibs enable-able from the refreshVersions Gradle extension

LouisCAD:


plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/RefreshVersionsExtension.kt
```diff
@@ -8,6 +8,11 @@ open class RefreshVersionsExtension {
 
     var versionsPropertiesFile: File? = null
     var extraArtifactVersionKeyRules: List<String> = emptyList()
+    var isBuildSrcLibsEnabled = false
```
jmfayard, line 11: Do we need to expose this variable in the DSL? `enableBuildSrcLibs` seems enough to me, I don't see the point of configuring `enableBuildSrcLibs = false`

jmfayard (approved):
What about the documentation? Is it up to date?

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/RefreshVersionsExtension.kt
LouisCAD: I can make it internal but cannot remove it entirely since the design of Gradle extensions require to have a backing field to be able to access what's been configured layer on.

LouisCAD:
The documentation hasn't been updated, and there's no KDoc for the extension yet, but for the case of buildSrcLibs itself, I think a while ago, we agreed on including it as part of a release first, and document it in a later release.

plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/RefreshVersionsExtension.kt
LouisCAD: Made `isBuildSrcLibsEnabled` internal.

Issue #343, opened 2021-05-08
Sometimes hundreds of dev versions are proposed by refreshVersions


jmfayard:
See for example inside http4k

> version.com.graphql-java..graphql-java=16.2

https://github.com/http4k/http4k/blob/380ec8a26cea303adeec7cde23e051caeb329dcf/versions.properties#L40-L575

vlsi:
```
version.commons-collections..commons-collections=3.2.1
##                                   # available=3.2.2
##                                   # available=20030418.083655
##                                   # available=20031027.000000
##                                   # available=20040102.233541
##                                   # available=20040616
```

```
version.commons-pool..commons-pool=1.6
##                     # available=20020501
##                     # available=20030623.172700
##                     # available=20030818.195203
##                     # available=20030821.192746
##                     # available=20030825.183949
```

What do you think if the plugin ignores versions released more than one year than the currently used version?

For instance, [commons-pool versions](https://search.maven.org/artifact/commons-pool/commons-pool) are as follows:

1.4 -- 16-Jan-2008
‎1.0.1 -- 24-Nov-2005
20030623.172700 -- 09-Nov-2005
20030825.183949 -- 09-Nov-2005

WDYT?

LouisCAD:
I don't think the release dates or timestamps are part of the maven metadata that we can get from repositories.

I might be wrong, but I recall trying to find such a thing twice, with no success.

vlsi:
You are right, the timestamp is not a part of metadata: https://repo1.maven.org/maven2/commons-pool/commons-pool/maven-metadata.xml :(

How about creating a curated list of exclusions then?

LouisCAD:
That's what I plan to do. I wish versioning was more conventional and strict through the Maven ecosystem.

jmfayard:
This will be solved by `rejectVersionsIf`

https://github.com/jmfayard/refreshVersions/pull/347

PR #347, opened 2021-05-13
API refreshVersions.rejectVersionIf { ... }

jmfayard:
About #343
Sometimes hundreds of dev versions are proposed by refreshVersions

It's what I had implemented in the Gradle Versions plugin
and it works well there

https://github.com/ben-manes/gradle-versions-plugin


LouisCAD:
I don't think this is the right design route because it makes it easy to filter all dependencies, potentially throwing the baby out with the bath water.

Also, regarding the example, it's important to note that we have a built-in way to know the stability level of a version.

Let's talk about it later and write all the options we could provide so we make.

jmfayard:
Yup, let's talk about it and see what are the use cases for such a feature

jmfayard:
@LouisCAD have a look at the updated PR, I improved the design and find it both simple and powerful
The filter is now not only declared but also implemented
I've tried it out on the kotlin-libraries-playground  https://github.com/LouisCAD/kotlin-libraries-playground/commit/e3838f16bb0cda35caadad030b8198742ff12bfb

LouisCAD:
One thing we could do is automatically add a comment in the `versions.properties` when there's code affecting the versions "acceptance".

For example:
```properties
## NOTE: Some versions are filtered by the rejectVersionsIf predicate. See the settings.gradle.kts file.
```

The `.kts` part would be dropped if we see the settings file is using Groovy DSL.

That way, people won't be surprised about versions not showing up, even if they just landed in the project.

jmfayard:
I'm fine with that!

On July 29, 2021, Louis CAD ***@***.***> wrote:
> One thing we could do is automatically add a comment in the
> versions.properties when there's code affecting the versions
> "acceptance".
>
> For example:
>
> ## NOTE: Some versions are filtered by the rejectVersionsIf predicate.
> See the settings.gradle.kts file.
> The .kts part would be dropped if we see the settings file is using
> Groovy DSL.
>
> That way, people won't be surprised about versions not showing up,
> even if they just landed in the project.
>

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/jmfayard/refreshVersions/pull/347#issuecomment-
889228668>, or unsubscribe
<https://github.com/notifications/unsubscribe-
auth/AADQFSHXZINZW7D22LAFSD3T2FVSJANCNFSM443KE73Q>.

jmfayard:
@LouisCAD Thinking about it it might be better to put the warning in the console log instead of the header.
As you said the header is only read once when it changes. If multiple people work on a project, only one will see the change, versus everyone if it's in the console log.

Also I'm pretty sure I can give the number of filtered versions instead of saying "some versions"

`NOTE: 172 versions are filtered by the rejectVersionsIf predicate. See the settings.gradle.kts file.`

LouisCAD:
@jmfayard I didn't say only one person would see it, I said once you see it in a project, you ignore it for next times.

I also think logs are more likely to be ignored than a comment, which can also be seen into the git diff.

I think adding a comment as I initially suggested is very important, and an informational or warning log could be a good addition if it tells how many version entries are affected, and how many newer versions are filtered out (as you implied with your example). That'd make it easy to spot if the predicate affects more version entries than expected.

jmfayard:
@LouisCAD I've implemented both the file comment and the log comment, ready to review

jmfayard:
I added the docs, this is ready to merge IHMO @LouisCAD

PR #353, opened 2021-05-17
New dependency notation base API

### LINKED ISSUE #352: Discuss/design how to contribute a dependency notation
Also revisit BOM support

jmfayard:
Closes #352 

See dependencies suggestions https://github.com/jmfayard/refreshVersions/issues/238


jmfayard:
@LouisCAD I've done the changes we discussed today, please have a look

jmfayard:
@LouisCAD I have updated the branch with the new API `DependencyGroup` and its use in `Spring, Http4k, Kodein, Orchid` 

It would be nice if you can prioritize reviewing this pull request, it has diverged from main since too much time, and it would be much easier to accept pull requests like #387 #389 once it's merged

plugins/dependencies/src/main/kotlin/dependencies/Http4k.kt
```diff
@@ -4,103 +4,106 @@ private val GROUP = "org.http4k"
...
-    val incubator = module("http4k-incubator")
-    val jsonrpc = module("http4k-jsonrpc")
-    val metricsMicrometer = module("http4k-metrics-micrometer")
-    val multipart = module("http4k-multipart")
-    val opentelemetry = module("http4k-opentelemetry")
-    val realtimeCore = module("http4k-realtime-core")
-    val resilience4j = module("http4k-resilience4j")
-    val securityOauth = module("http4k-security-oauth")
+    val bom: String get() {
+        usePlatformConstraints = true
```
LouisCAD, line 14: Nice.

plugins/dependencies/src/main/kotlin/dependencies/Http4k.kt
```diff
@@ -4,103 +4,106 @@ private val GROUP = "org.http4k"
...
-        val jetty = module("http4k-client-jetty")
-        val okhttp = module("http4k-client-okhttp")
-        val websocket = module("http4k-client-websocket")
+        val apache get() = module("http4k-client-apache")
+        val apacheAsync get() = module("http4k-client-apache-async")
+        val apache4 get() = module("http4k-client-apache4")
+        val apache4Async get() = module("http4k-client-apache4-async")
+        val jetty get() = module("http4k-client-jetty")
+        val okhttp get() = module("http4k-client-okhttp")
+        val websocket get() = module("http4k-client-websocket")
```
LouisCAD, line 41: We could also use property delegation, with inline getters and setters to allow the `KProperty` to be optimized out once we compile against Kotlin 1.4 (and drop support for Gradle 6).

plugins/dependencies/src/main/kotlin/dependencies/Http4k.kt
LouisCAD: Now that I'm thinking about it again…
I think it won't be that easy: in this form, one module could affect another, which can lead to surprising results that depend on non deterministic execution order.

We would want to attach the BoM activation on a per configuration, or a per module basis. Not sure how we can do that at the moment though.

PR #354, opened 2021-05-18
Add refreshVersionsCleanup task

jmfayard:
Remove the comments from versions.properties


LouisCAD:
Don't you think that a `refreshVersionsCleanup` dedicated task would make more sense?

First the `refreshVersions` task would not do something completely different because of an option.
Second, it'd better surface in the IDE.
Third, no risk of the option being added and not being removed in the run configuration, leading to user confusion when you they actually want to refresh the versions.

jmfayard:
I created a group "refreshVersions" for our tasks, we already have 6 of them and there are more coming

PR #362, opened 2021-05-29
:refreshVersionsMissingEntries allows to migrate an existing projet

jmfayard:
## 🚀 Description

Add a task refreshVersionsMissingEntries that populates the versions.properties file with all hardcoded versions, including plugins.

This is mostly a refactoring of something that was done inside :buildSrcLibs  

I have a magic regex to make the build use the placeholder version here ==> https://regexr.com/5ti1j

It was tested in projects like

- https://github.com/kotest/kotest/pull/2280
- https://github.com/necatisozer/kotlin-analytics/pull/1
- https://github.com/pavlospt/refactored-umbrella/pull/21
- https://github.com/kittinunf/fuel/pull/811

## ✅ Checklist


LouisCAD (changes requested):
The changes to the issue templates needs be done directly on the main branch, or in a dedicated PR, because they are not related in any way to the title of this PR.
Same for the removal of the PR template.

There's currently a bigger problem in this PR: it reverts the lint updater code and all the tests.

Please, ensure that the diff is only what makes the refreshVersionsMissingEntries work and tested.

Also, I don't think the changes to the `versions.properties` file of the `sample-multi-modules` are necessary or needed for this PR.

Finally, I see that there's the version catalogs feature enabled for that same sample, but I don't think it should be in this PR either.

As you can see, I'm up late, which mean I'll not be able to get back to this for 9 hours at the very least (so 1PM+), plus you know I work on another project. Consequently, if you want this to be merged rather sooner than later, try to make this PR as perfect as you can before re-requesting my review 🙂

jmfayard:
@LouisCAD sorry for the git mess. It should be all good now. Famous last words.

PR #366, opened 2021-06-04
NPM support

NikkyAI:
## 🚀 Description

adds a Resolver for NPM dependencies and required hooks to make use of it

## 📄 Motivation and Context
multipltform!!

tested with `:sample-kotlin-js` (some dependencies copied from a fritz2 webapp)

## 📦 Types of changes

## ✅ Checklist

not specifying a version in `npm("dpendency")` sets it to `"*"` for my own projects i would probably want to add this as a version placeholder, since i never want to not specfy a version
maybe that could be a feature flag ?

we might want to look into adding special handling for npm version ranges.. as most apparently use `^1.0.0` rather than `1.0.0` currently refreshVersions decides this is smaller than any other version.. so it lists all available versions, which suits me well.. but not sure if that is how it is supposed to go

so far this is just hacked in with the least amount of changes in the code
it would make sense to refactor that, ... maybe split ModuleIdentier into a selaed class for Maven and Npm identiers


LouisCAD:
@NikkyAI Can you fix what breaks in the CI using Linux or macOS?

Also, are there any concerns beyond this that you are aware of?
I'm thinking about merging that relatively soon.

LouisCAD:
Thank you very much for your work Nikky!
Will be in next release!

Issue #205, opened 2020-07-13
Add first-class support for included/composite builds


LouisCAD:
Right now, included builds are treated like regular projects, which means their versions are separate from the the project that includes them.

While this is great when you want to keep the projects decoupled, it'd be great to also have the option to share the setup, the `versions.properties` file, and have runs of the `refreshVersion` task also refresh the versions from all included builds that opt-in to this behavior.

Savrov:
Hello, you asked us (developers) to describe here use cases, where we would need that feature.  So, I have a Kotlin MPP, that has multiple modules:
- common
- backend
- frontend

So, it will be great if we could update `versions.properties` to manage dependencies in an entire project.

LouisCAD:
@Savrov This is already supported.

This issue is not about multi-module projects but about `includedBuilds`support, which is different.

NikkyAI:
do included builds work at all when refreshVersions is involved?
i do not need the versions.properties to be synced or shared..

when i try to run the task in the parent project i get errors like

```
* What went wrong:
Execution failed for task ':refreshVersions'.
> Couldn't resolve version for ModuleId(group=org.springframework, name=spring-context-indexer)
```

and idea gradle sync shows this:

```
Property versionsPropertiesFile not initialized yet! Has it been used after reset or before init?
```

oblitzitate:
I saw the following articles that convinced to prefer composite builds over buildSrc:

[Stop using Gradle buildSrc. Use composite builds instead](https://proandroiddev.com/stop-using-gradle-buildsrc-use-composite-builds-instead-3c38ac7a2ab3)
[How to use Composite builds as a replacement of buildSrc in Gradle](https://medium.com/bumble-tech/how-to-use-composite-builds-as-a-replacement-of-buildsrc-in-gradle-64ff99344b58)

According to the articles, buildSrc is inefficient compared to composite builds so I'm definitely all for first-class support to composite builds.

NikkyAI:
in the newest versions of gradle buildSrc is a included build, so not sure how much those articles still hold true

NikkyAI:
**dirty** experiment to see if refreshVersions can automatically set up task dependencies on refreshVersions in included builds

```
gradle.rootProject {
    afterEvaluate {
        val task = gradle.includedBuilds.firstOrNull()?.task(":refreshVersionsDDDDsjdfhdsiuf")!!
        logger.lifecycle("task: $task")
[...]

PR #367, opened 2021-06-05
fix idea sync failing on composite builds, add support for composite builds

NikkyAI:
## 🚀 Description

handles failure due to uninitialized `versionPropertiesFile` during idea sync

it does this by catching the exception and setting the description of task `refrewshVersions` to the most likely correct value

main project and included project will have seperate `versions.properties`
(although if they both are configured to point at the same file it might work?)

it also refactors `RefreshVersionsConfigHolder` usage
using the rootProject path as the key each uses their own `RefresVersionsConfig` instance
this avoids the parentProject writing to the included builds `versions.properties`

the only breaking change is for groovy code: `versionFor` needs to be passed the project to grab the correct config instance
(works out of the box as receiver in kotlin)
unless there is a way to make these functions have a receiver without breaking usage code ?

`versionFor` was moved into the extension `versions`, so that it can keep a reference to the correct config
the usage looks like so now: 
```
version.versionFor("version.kotlin")

// alternative

versions {
	versionFor("version.kotlin")
}
```
it does not require imports in the buildscript anymore

## 📄 Motivation and Context

makes `includeBuild` not fail
this mainly fixes the idea sync failing and removing all tasks from the tool window

should unblock work on https://github.com/jmfayard/refreshVersions/issues/205

tested on idea ultimate on windows and linux
all sample projects pass checks and idea sync

## 📦 Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## ✅ Checklist

- [x] I have updated the documentation accordingly.


LouisCAD:
I've seen I broke something with Kotlin plugin version by trying to fix something, I'll fix it on the main branch first.

LouisCAD:
@NikkyAI I fixed what I broke on the main branch, you can undo the changes you did to work that around in this PR.

NikkyAI:
@LouisCAD i reverted the commit, kepts the plugin alias so that tests do not fail, i will remove that in a future PR when i work on integrating parent and included builds more

LouisCAD:
If you merge latest main again, the issue will go away.

NikkyAI:
further testing reveals this breaks in different ways.. still needs more testing

PR #369, opened 2021-06-07
Tests: make BundledDependenciesTest incremental

jmfayard:


plugins/dependencies/src/test/kotlin/de/fayard/refreshVersions/rules/BundledDependenciesTest.kt
```diff
@@ -1,59 +1,73 @@
...
+                            dependencyMapping to version
+                        }
+                    }
+                }.awaitAll()
+        }
+
+        val newValidatesMappings = mappings
+            .filter { it.second != null }
+            .joinToString(prefix = "\n", separator = "\n") { (mapping, version) ->
+                "${mapping.group}:${mapping.artifact}:${version!!.value}"
```
LouisCAD, line 64: Why keep the version? I think we only need to know whether the given maven coordinates are on the repo or not.

jmfayard: It's not strictly necessary but it's a useful hint that we have actually parsed the pom.xml, and may help to see whether all versions are aligned - and a rule is needed - or not @louiscad

LouisCAD: Well, alignment of version is out of the scope of this test IMO. It's something that we manually check when we add a dependency, and something that we could automatically infer separately.

I am concerned that the tested artifacts list, if kept as is, would have outdated versions, unless we make the test check everything every time, which would incur a penalty for each run of all tests (over one minute over a 1Gbps fiber connection from my experience).

Therefore, I think it's best to not look and keep the version at all here.

jmfayard: I removed the versions

PR #383, opened 2021-06-29
Add a tl;dr in the readme

jmfayard:
Say just enough to be useful,
but not enough to not give the impression that there is no need to read the docs.


LouisCAD:
We need to publish the release with migration facilities first. Will try to do it this Wednesday.

PR #399, opened 2021-08-06
refreshVersionsMigrate use dependency constant names when available

jmfayard:
## 🚀 Description
One part we left out in `refreshVersionsMigrate` was to use the replace the maven coordinates with the matching dependency notation, if available.

Turns out it's possible to do it.

## 📄 Motivation and Context
Better migration

https://github.com/jmfayard/ktor-samples/commit/e101b1cc00e31b4f271358ad199a15e18518655a

https://github.com/jmfayard/Anki-Android/commit/902edb9d203fe9c2e85572133b890a133c8d713e

## 📦 Types of changes

## ✅ Checklist

- [ ] My change requires a change to the documentation.


jmfayard:
The check failed is because `markDependencyNotationsUsage() ` doesn't make sense if we use all dependencies via `getArtifactNameToConstantMapping()`

LouisCAD:
I knew it was possible and I told you 🙃

LouisCAD:
Let's find a way to disable those checks while we are getting the mapping, I already have an idea on how we can do it.

jmfayard:
Done, ready to review.
The diff seems big but it's mostly that `RefreshVersionsMigrateTask` and its tests had to be moved from core to dependencies

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/DependencyGroup.kt
```diff
@@ -27,6 +27,7 @@ open class DependencyGroup(
         val ALL_RULES: List<ArtifactVersionKeyRule>
             get() = ALL.mapNotNull { it.rule }
 
+        var disableBomCheck: Boolean = false
```
LouisCAD, line 30: Shouldn't this property be `internal`, or `@InternalRefreshVersionsApi`?

jmfayard: done, ,thanks

PR #408, opened 2021-08-15
Firebase dependency notations upgrades

LouisCAD:
This PR updates and cleans-up a bit the Firebase Dependency notations.

It also adds a way to use or not use platform constraints (e.g. from a BoM) for a given dependency notation, or specify a specific version, which can be handy when you need to deviate from the version from `versions.properties` for a given module. An example use case would be supporting Android 4, and sticking to OkHttp 3.12 for that module, while using the latest for other modules.

Here's the accepted syntax, where the latter only works for Kotlin DSL:

```kts
implementation(Firebase.cloudMessagingKtx.withVersionPlaceholder())
implementation(Firebase.cloudMessagingKtx.withoutVersion())
implementation(Firebase.cloudMessagingKtx.withVersion("0.1.2"))

implementation(Firebase.cloudMessagingKtx(version = "0.1.2"))
implementation(Firebase.cloudMessagingKtx(version = "_"))
implementation(Firebase.cloudMessagingKtx(version = null))
```

Here are all the APIs that we've considered before committing to one:

```kts
implementation(Firebase.cloudMessagingKtx.withoutPlatformConstraints())
implementation(Firebase.cloudMessagingKtx.withPlatformConstraints())
implementation(Firebase.cloudMessagingKtx(usePlatformConstraints = false))
implementation(Firebase.cloudMessagingKtx(withVersionPlaceholder = false))

implementation(Firebase.cloudMessagingKtx.withVersionPlaceholder())
implementation(Firebase.cloudMessagingKtx.withoutVersion())
implementation(Firebase.cloudMessagingKtx.withVersion("0.1.2"))

implementation(Firebase.cloudMessagingKtx(version = "0.1.2"))
implementation(Firebase.cloudMessagingKtx(version = "_"))
implementation(Firebase.cloudMessagingKtx(version = null))

implementation(Firebase.cloudMessagingKtx.withVersion("0.1.2"))
implementation(Firebase.cloudMessagingKtx.withVersion("_"))
implementation(Firebase.cloudMessagingKtx.withVersion(null))

implementation(Firebase.cloudMessagingKtx.withVersion(null))
implementation(Firebase.cloudMessagingKtx.withoutPlatformConstraints())
implementation(Firebase.cloudMessagingKtx.bom)
implementation(Firebase.cloudMessagingKtx.bom())
implementation(Firebase.cloudMessagingKtx.noBom)
implementation(Firebase.cloudMessagingKtx.noBom())
implementation(Firebase.cloudMessagingKtx.noVersion)
implementation(Firebase.cloudMessagingKtx.withVersion)
implementation(Firebase.cloudMessagingKtx.withoutVersion)
implementation(Firebase.cloudMessagingKtx.withoutVersion())
implementation(Firebase.cloudMessagingKtx(withVersion = false))
```


LouisCAD:
Merging, because I believe it's already an improvement over what we had before.
However, reviews are still appreciated and can be addressed in further commits or PRs.

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/DependencyGroup.kt
```diff
@@ -0,0 +1,172 @@
...
+) : DependencyNotation {
+
+    protected abstract fun shouldUsePlatformConstraints(): Boolean
+
+    override fun withVersionPlaceholder(): String = this(version = "_")
+    override fun withVersion(version: String): String = this(version = version)
+    override fun withoutVersion(): String = this(version = null)
+
+    override operator fun invoke(version: String?): String = when (version) {
+        null -> artifactPrefix
```
jmfayard, line 61: We can use `.isNullOrBlank()` instead, for the least surprising behavior in case `""` or `"   "` are passed.

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/DependencyGroup.kt
```diff
@@ -0,0 +1,172 @@
...
+                            "You are trying to use a BoM ($artifactPrefix), " +
+                                "but dependency notations relying on it have been declared before! " +
+                                "Declare the BoM first to fix this issue."
+                        )
+                    }
+                    this@AbstractDependencyGroup.usePlatformConstraints = true
+                } else if (isGroupUsingPlatformConstraints) {
+                    haveDependencyNotationsWithPlatformConstraintsBeenUsed = true
+                }
+            }
```
jmfayard, line 167: Please, double check that the boolean logic is sound

LouisCAD: I think the issue is that the naming `usePlatformConstraints` is not very easy to hold in mind as to what it actually means.

I'll try to find something that is easier to reason about and check that logic.

LouisCAD: Resolved on main

PR #409, opened 2021-08-15
Cleanup buildSrcLibs

LouisCAD:
I moved 2 files that didn't need to be in the core module, and I could make many, many symbols `internal` instead of `public` + `@InternalRefreshVersionsApi`. I don't get why they've been put there. Even the package header of the `DependencyNotations.kt` file was suggesting it was meant to be in the `buildSrcLibs` module.


jmfayard:
I think they were needed for `./gradlew refreshVersionsCatalog`

jmfayard (approved):
approved

PR #410, opened 2021-08-19
Refactor more dependency notations to the new DependencyGroup API

jmfayard:


plugins/dependencies/src/test/kotlin/de/fayard/NonRegression.kt
```diff
@@ -16,6 +18,17 @@ val testResources: File = File(".").absoluteFile.resolve("src/test/resources")
...
+        ALL_DEPENDENCIES_NOTATIONS
+        val rulesDir = File(".").absoluteFile.resolve("src/main/resources/refreshVersions-rules")
+        val file = rulesDir.resolve("dependency-groups-alias-rules.txt")
+        val content = AbstractDependencyGroup.ALL_RULES
+            .sorted()
+            .distinct()
+            .joinToString(separator = "\n\n")
+        file.writeText(content)
+    }
+
```
LouisCAD, line 31: @jmfayard Why did you add that?

LouisCAD: Oh wait, I see that you actually moved it from one file to another. Still wondering why though.

LouisCAD: I mean, this test is not about non regression. Maybe we should rename this file to something like `BundledDependencyNotationsTests`?

jmfayard: @LouisCAD I only moved it from another test because it belongs together

PR #411, opened 2021-08-19
Detection of BoM usage doesn't work

jmfayard:
Now that the creation of DependencyNotation-s is not lazy, the detection of BoM usage doesn't work. 

The following change to build.gradle.kts should throw an error  but doens't

### DIFF
--- a/sample-kotlin/build.gradle.kts
+++ b/sample-kotlin/build.gradle.kts
@@ -27,6 +27,10 @@ repositories {
 }
 
 dependencies {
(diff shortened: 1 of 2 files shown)


LouisCAD:
@jmfayard The title of the commit and of the PR is what should be the title of the issue.

LouisCAD:
We found out that it's actually expected, but we might come up with a way to enforce usage of the built-in bom. Closing.

PR #421, opened 2021-08-24
[Issue-238] Add Koin

brady-aiello:
This addresses the Koin part of this issue: https://github.com/jmfayard/refreshVersions/issues/238

## 🚀 Description
Add the Koin dependencies. [Koin docs](https://insert-koin.io/docs/setup/v3)

## 📄 Motivation and Context
Koin is a common DI / Service Locator framework for KMP, and we'd like
to migrate to refreshVersions completely.
https://insert-koin.io/docs/setup/v3
How we're using it for safe multithreading on native in KaMP Kit:
https://github.com/touchlab/KaMPKit/blob/main/shared/src/commonMain/kotlin/co/touchlab/kampkit/Koin.kt

- `./gradlew :refreshVersions:build`
- verified that the coordinates are generated in
  `dependencies-mapping-validated.txt` and
`bundled-dependencies-validated.txt`.

## 📦 Types of changes

- [X] New feature (non-breaking change which adds functionality)

## ✅ Checklist

- [X] I have read the [guidelines for the development process](https://jmfayard.github.io/refreshVersions/contributing/submitting-prs/dev-process/)
- [ ] My change requires a change to the documentation.


plugins/dependencies/src/main/kotlin/dependencies/Koin.kt
```diff
@@ -0,0 +1,28 @@
...
+ * A pragmatic lightweight dependency injection framework for Kotlin developers.
+ * Koin is a DSL, a light container and a pragmatic API
+ *
+ * - [Official website here](https://insert-koin.io/)
+ * - GitHub page: [Koin](https://github.com/InsertKoinIO/koin)
+ * - [GitHub Releases here](https://github.com/InsertKoinIO/koin/releases)
+ */
+@Incubating
+object Koin : DependencyGroup(
+    group = "io.insert-koin"
```
jmfayard, line 16: Don't all Koin dependencies share the same version?
If yes, you should define here a `rawRule` similar to the ones here https://github.com/jmfayard/refreshVersions/tree/main/plugins/dependencies/src/main/resources/refreshVersions-rules

The file  `plugins/dependencies/src/test/resources/dependencies-versions-key-validated.txt` should then show `version.koin` instead of `version.NO-RULE`

brady-aiello: Will do, thanks!

LouisCAD:
@jmfayard Do you think we should keep the key as `insert-koin`, or should we use `koin` instead?

plugins/dependencies/src/main/kotlin/dependencies/Koin.kt
jmfayard: Would you mind changing the rule so that the key is `version.koin` insert of `version.insert-koin`?
Apart from that great job!

jmfayard:
Merged, thanks a lot @brady-aiello

brady-aiello:
Sweet! Thank you!

PR #446, opened 2021-09-28
Kotest dependencies fixes

Kantis:
## 🚀 Description
1. Kotest extensions have independent lifecycles. Meaning the latest version of `kotest-extensions-spring` might be 2.0.0 while `kotest-extensions-koin` might still be at `1.1.0`. Currently, these are bundled into a single versioning property.
2. Added notation for two more kotest dependencies (`Testing.kotest.framework.api` for `kotest-framework-api` and `Testing.kotest.framework.datatest` for `kotest-framework-datatest`)

Please let me know if I missed something.


plugins/dependencies/src/main/resources/refreshVersions-rules/testing-version-alias-rules.txt
```diff
@@ -7,9 +7,6 @@ org.junit.jupiter:junit-jupiter(-*)
 io.kotest:*
    ^^^^^^
 
-io.kotest.extensions:kotest-extensions-*
-                     ^^^^^^^^^^^^^^^^^
-
```
LouisCAD, line 12: This should not be removed, but edited instead so it resolves to `version.kotest`.

plugins/dependencies/src/test/resources/dependencies-versions-key-validated.txt
```diff
@@ -603,21 +603,23 @@ io.kotest..kotest-extensions-koin=version.kotest
...
+io.kotest.extensions..kotest-extensions-allure=version.NO-RULE
+io.kotest.extensions..kotest-extensions-embedded-kafka=version.NO-RULE
+io.kotest.extensions..kotest-extensions-gherkin=version.NO-RULE
+io.kotest.extensions..kotest-extensions-koin=version.NO-RULE
+io.kotest.extensions..kotest-extensions-mockserver=version.NO-RULE
+io.kotest.extensions..kotest-extensions-pitest=version.NO-RULE
+io.kotest.extensions..kotest-extensions-robolectric=version.NO-RULE
+io.kotest.extensions..kotest-extensions-spring=version.NO-RULE
+io.kotest.extensions..kotest-extensions-testcontainers=version.NO-RULE
+io.kotest.extensions..kotest-extensions-wiremock=version.NO-RULE
```
LouisCAD, line 622: As you can see, there's no rule remaining here, which means it would default back to the longer `version.maven-group..name` syntax, defeating the purpose of your intention in this PR.

LouisCAD (changes requested):
Thanks for the PR.
There's an issue regarding your version key rules changes.
I'd recommend to rollback the changes into the txt files before implementing the fix and running the tests again.

plugins/dependencies/src/test/resources/dependencies-versions-key-validated.txt
Kantis: That's the point of the PR. `io.kotest.extensions:kotest-extensions-wiremock` is not tied to any other extension's lifecycle, nor the main kotest lifecycle. So it should be treated independently

Kantis: The other alternative would be for each extension to resolve to a nicer name, like `version.kotest.extension.allure`, `version.kotest.extension.spring`, etc.

LouisCAD: I think I understand. Every extension is supposed to have a separate, independent version, is that right?

In that case, yes to adding a rule that will make the version key a bit nicer.

LouisCAD:
There's no need for one rule per artifact. You can use stars, see the examples in other rules already in refreshVersions (linked from the contributing docs).

Kantis:
Ah, thank you! That's much nicer 😃

LouisCAD:
Thanks for the contribution!
Will check it locally soon.

LouisCAD:
It looks like the PR is not based on the main branch from 2 days ago as the PR submission would suggest, but it based on a commit from over 3 weeks ago 🤔

LouisCAD:
Also looks like you initially, correctly commited the `removed-dependencies-versions-keys.txt` file, but then wrongly removed it, while it's ciritical to have it to ensure automatic migration to the new version key, without upgrading anything.

LouisCAD (approved):
Fixed what needed to be fixed.

Thanks!

PR #455, opened 2021-11-03
Dependency notations auto removals

LouisCAD:
## 🚀 Description

Add the ability for refreshVersions to automatically replace previously built-in dependency notations that have been removed with the maven coordinates or any new replacement with the exact same coordinates, and suggest any dependency replacement that might be needed (e.g. when coordinates change).

## 📄 Motivation and Context

Google and other community members often deprecate and/or replace dependencies, and we want to get rid of them as well, without breaking refreshVersions users, while helping them migrate.

There are several new unit tests.

As for integration tests:
Fictive future versions (with different maven coordinates and plugin ids to avoid clash) have been tested, to ensure both snapshots and releases are handled correctly, as desired for this dependency notations removal process.

## 📦 Type of changes

## ✅ Checklist
- [ ] I have updated the documentation accordingly. (Will be done along with the change log update)


LouisCAD:
The fact that Gradle 6.8+ is required now appears correctly in the error message.

LouisCAD:
Offline tests where snapshot and releases have been simulated have been successful.

Finally ready to merge! 🥳

PR #494, opened 2022-01-22
Error tolerance

### LINKED ISSUE #468: Don't fail completely if a dependency's version info can't be found
- [X] I have read the [guidelines for submitting issues](https://jmfayard.github.io/refreshVersions/contributing/submitting-issues/)

## ⚠️ Is your feature request related to a problem? Please describe
An internal project owned by another team is not published correctly, so `refreshVersions` can't read its version. But when I try to run `refreshVersions`, it errors out and doesn't report any new versions. So my only choice is to either hard-code the version in `build.gradle`, or comment out the offending project so that the error doesn't happen. The first means all my versions wouldn't be in the same location. The second is a lot of fiddling every time I want to check versions.

## 💡 Describe the solution you'd like
If `refreshVersions` can't find a dependency, it should just put a comment (e.g., `# no version info could be found`) and add all the version info for the other dependencies, rather than erroring and not updating any of the versions.

## 🤚 Do you want to develop this feature yourself?

- [ ] Yes
- [X] No
but I'd be happy to. I suspect the hardest part would be figuring out where all this happens.

LouisCAD:
## 🚀 Description

These changes avoid refreshVersions crashing if there's one failure for one of the repos or dependencies, or related to the network stability. Instead, error messages will be inserted contextually in the `versions.properties` file.

## 📄 Motivation and Context

This PR resolves #468

Tested locally with purposefully wrong repository urls.
Worked well for timeout, and for non existing domain:
- failure comments are added at the right place in `versions.properties` if an error happens.
- failure comments are updated if the errors change.
- failure comments are removed after running `refreshVersions` if there's no errors in the subsequent run.

In the case of non-repo existing domains, they return 404 errors, and we can't differenciated them from what an actual repo missing the requested artifact would return, so they are ignored in both cases.

### DIFF
--- a/plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/DependencyVersionsFetcher.kt
+++ b/plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/DependencyVersionsFetcher.kt
@@ -1,37 +1,69 @@
 package de.fayard.refreshVersions.core
 
+import okio.IOException
 import org.gradle.api.Incubating
-import java.io.IOException
 
 @Incubating
 abstract class DependencyVersionsFetcher protected constructor(
     val moduleId: ModuleId,
-    val repoKey: Any
+    val repoUrlOrKey: String
 ) {
 
     companion object;
 
-    @Throws(IOException::class, Exception::class)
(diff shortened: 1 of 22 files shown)


LouisCAD:
Let's see

PR #496, opened 2022-01-23
Remove LegacyBootstrapUpdater and LegacyBootstrapMigrator

LouisCAD:
## 🚀 Description

These are no longer needed since refreshVersions has now been a settings plugin
for a while, no longer needing the special bootstrap we used to have, and all
our users should have upgraded by now, and if not, they can just upgrade to
a previous version before to not have their build break.

The code removed in this commit can still serve for inspiration purposes for code
that would need to update other code, especially when these are Gradle files.

## 📄 Motivation and Context

Cleaning up dead code.

CI will do it.


LouisCAD:
Testing something

PR #537, opened 2022-04-14
Change AGP to dependency notation

iainism:
Hello,

I spotted that the Android Sample App uses the underscore notation instead of dependency notation for the AGP.

This changes the build.gradle.kts to use dependency notation instead.

## Why?

Updating of example

## How?

Edited build.gradle.kts to use dependency notation for AGP

## Testing?

Built project with no errors/warnings after change.

### DIFF
--- a/sample-android/build.gradle.kts
+++ b/sample-android/build.gradle.kts
@@ -8,7 +8,7 @@ buildscript {
   }
 
   dependencies {
-    classpath("com.android.tools.build:gradle:_")
+    classpath(Android.tools.build.gradlePlugin)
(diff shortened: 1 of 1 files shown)


iainism:
I haven't yet used the Kotlin DSL in Gradle files so I'm not sure I can contribute too many further improvements!

(I was actually looking to see if there was a dependency notation possible in place of `"org.jetbrains.kotlin:kotlin-gradle-plugin:_"` after seeing 'Android.tools.build.gradlePlugin` had been substituted by auto-migration in one of my one build.gradle files, and I worked out that it could be changed in the sample)

jmfayard:
Merged, thanks @iainism

PR #546, opened 2022-05-28
Memory leak checks

### LINKED ISSUE #536: Memory leak
- [x] I have read the [guidelines for submitting issues](https://jmfayard.github.io/refreshVersions/contributing/submitting-issues/)
## 🐛 Describe the bug
When using the version 0.40.1, our builds fail with the following error
```
an FAILURE: Build failed with an exception.
* What went wrong:
GC overhead limit exceeded
```

After adding `+HeapDumpOnOutOfMemoryError` to the jvmargs and analysing the hprof we got the following report from [Eclipse MAT](https://www.eclipse.org/mat/) 

Screenshot 2022-04-13 at 16 44 10

Note : 
Current jvmargs in our gradle.properties
`org.gradle.jvmargs=-Xmx4096m -XX:+HeapDumpOnOutOfMemoryError`

## ⚠️ Current behavior

After multiple builds we have to restart the gradle deamon because it is taking most of the allowed allocated memory.
Not stopping it result in very slow builds and the above mentioned error.

## ✅ Expected behavior

That our build time remains consistent

## 💣 Steps to reproduce

N/A

## 📱 Tech info
N/A

LouisCAD:
Hello,

## What and How?

I'm adding a check for memory leak in the `check-build.yml` workflow. It will use the `sample-groovy` project.
When limiting the JVM heap to 50MB, just 2 runs of `./gradlew help`/`.gradlew` are needed to trigger a JVM heap space exhaustion when a memory leak is present.

Fixes #536

## Why?

To detect memory leaks on the long term.


LouisCAD:
In-CI memory leak detection check is finally working after a lot of fine-tuning, trial, and error.

I now plan to tackle #536 (still in this PR) and be done with leaks.

LouisCAD:
This is ready to be merged. I'll do it tomorrow and make a release.

In addition to the CI check that tries to run the sample-groovy Gradle project 10 times in a row, I tested locally 399 subsequent runs, and it worked fine after the fix while I had put the following in the `gradle.properties` file:
```properties
org.gradle.jvmargs=-Xmx100M -XX:+HeapDumpOnOutOfMemoryError
```

### Screenshot:

Screenshot 2022-05-30 at 20 51 57

PR #550, opened 2022-06-01
Remove navigation-compose rule

joharei:
Hello/bonjour,

## What?

Removed the rule for the `navigation-compose` artifact.

## Why?

The versioning of this artifact [has been aligned](https://developer.android.com/jetpack/androidx/releases/navigation#navigation_compose_version_100_2) with the rest of `androidx.navigation`.

## How?

Just removed the rule and ran `./checkPlugins.sh`. Tell me if there is anything else that should be done.

## Testing?

Ran `./checkPlugins.sh`.


jmfayard:
Thanks @joharei 
I've merged it via https://github.com/jmfayard/refreshVersions/pull/559

PR #566, opened 2022-07-24
Complete OkHttp dependencies.

yschimke:
## What?

Adds missing OkHttp dependencies.

## Why?

OkHttp5 Alpha adds new dependencies, but also some were already missing from 4.

## How?

Review https://mvnrepository.com/artifact/com.squareup.okhttp3

## Testing?

Ran tests locally.


jmfayard:
I was wondering:

- whether those should be in an `OkHttp5` object, but they are still the okhttp3 maven group so I guess that's ok
- whether we should deprecate some dependency, but okhttp5 is not released as stable so that doesn't make sense.

In sum LGTM, merged & thanks 🙏🏻

LouisCAD:
Thank you for the contribution!

PR #569, opened 2022-08-02
Add AndroidX.lifecycle.runtimeCompose dependency notation

necatisozer:
Hello/bonjour,

## What?

Resolves #XX 

Tell us what these changes are doing.

## Why?

Tell us why those changes are needed.

## How?

Tell us briefly what you did here, and what your solution/implementation is doing.

## Testing?

Tell us how you tested these changes. We will test as well.


LouisCAD:
Hello, thanks for the heads up, I finally added it properly as I caught up with 3 months of AndroidX updates.
Will be included in the next release.

Issue #582, opened 2022-09-05
Support for Kotlin scripts


jmfayard:
I propose we add first-class support for [Kotlin scripts](https://kotlinlang.org/docs/custom-script-deps-tutorial.html) because life is too short for Bash programming. 

### FeatureFlag
`KOTLIN_SCRIPTS`, disabled by default

### refreshVersions

`./gradlew refreshVersions --enable KOTLIN_SCRIPTS` should show updates
Initial:
```
#!/usr/bin/env kotlin  
  
@file:Repository("https://jitpack.io")  
@file:DependsOn("io.rsocket.kotlin:rsocket-core-jvm:0.13.0-SNAPSHOT")  
@file:DependsOn("io.ktor:ktor-client-okhttp:1.5.2")  
@file:DependsOn("com.squareup.wire:wire-moshi-adapter:3.4.0")  
@file:CompilerOptions("-jvm-target", "17")
```

- `mavenCentral()` should be supported by default
- `@file:Repository(..)` should be supported if present

Expected:

```
#!/usr/bin/env kotlin  
  
@file:Repository("https://jitpack.io")  
@file:DependsOn("io.rsocket.kotlin:rsocket-core-jvm:0.13.0-SNAPSHOT")  
////                                    # available:0.15.0")  
////                                    # available:0.16.0")  
@file:DependsOn("io.ktor:ktor-client-okhttp:1.5.2")  
////                            # available:1.6.0")  
////                            # available:1.7.0")  
@file:DependsOn("com.squareup.wire:wire-moshi-adapter:3.4.0")  
@file:CompilerOptions("-jvm-target", "17")
```
 
### refreshVersionsCleanUp

`./gradlew refreshVersionsCleanUp --enable KOTLIN_SCRIPTS` should remove the comments

Initial:
```
#!/usr/bin/env kotlin  
  
@file:Repository("https://jcenter.bintray.com")  
//// somme comment
@file:DependsOn("io.rsocket.kotlin:rsocket-core-jvm:0.13.0-SNAPSHOT")  
////                                    # available:0.14.0")  
////                                    # available:0.15.0")  
@file:DependsOn("io.ktor:ktor-client-okhttp:1.5.2")  
```

Expected:
```
#!/usr/bin/env kotlin  
  
@file:Repository("https://jcenter.bintray.com")  
@file:DependsOn("io.rsocket.kotlin:rsocket-core-jvm:0.13.0-SNAPSHOT")  
[...]

PR #583, opened 2022-09-05
Add KOTLIN_SCRIPTS feature flag

jmfayard:
Minimum viable PR for #582

Discussion should happen in the issue


plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/RefreshVersionsCleanupTask.kt
```diff
@@ -59,4 +80,11 @@ open class RefreshVersionsCleanupTask : DefaultTask() {
             OutputFile.GRADLE_VERSIONS_CATALOG.logFileWasModified()
         }
     }
+
+    @TaskAction
+    fun cleanupKotlinScripts() {
+        if (FeatureFlag.KOTLIN_SCRIPTS.isEnabled) {
+            println("TODO: refreshVersionsCleanUp should clean up Kotlin Scripts see https://github.com/jmfayard/refreshVersions/issues/582")
```
LouisCAD, line 87:
```suggestion
            println("NOTE: refreshVersionsCleanUp doesn't clean up Kotlin Scripts yet, see https://github.com/jmfayard/refreshVersions/issues/582")
```

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/RefreshVersionsTask.kt
```diff
@@ -115,6 +115,9 @@ open class RefreshVersionsTask : DefaultTask() {
                 }
             }
         }
+        if (FeatureFlag.KOTLIN_SCRIPTS.isEnabled) {
+            println("TODO: refreshVersions should support Kotlin Scripts see https://github.com/jmfayard/refreshVersions/issues/582")
```
LouisCAD, line 119:
```suggestion
            println("NOTE: refreshVersions support for Kotlin Scripts isn't implemented yet, see https://github.com/jmfayard/refreshVersions/issues/582")
```

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/RefreshVersionsCleanupTask.kt
```diff
@@ -11,11 +11,32 @@ import de.fayard.refreshVersions.core.internal.versions.VersionsPropertiesModel.
...
+        }
+
+    @Input
+    @Optional
+    @Option(option = "disable", description = "Disable a feature flag")
+    var disableFlag: FeatureFlag? = null
+        set(value) {
+            field = value
+            if (value != null) FeatureFlag.userSettings.put(value, false)
+        }
```
LouisCAD, line 38: I think this should be in a separate PR, don't you?

LouisCAD (commented):
A few inline comments.

Also, a general thought I had for a while: We probably want to not depend on being a Gradle plugin for Kotlin Scripts support.
That means potentially a lot of refactoring to make a Gradle independent core.
We can sure start with a version that depends on the Gradle plugin and a host Gradle project to run, but that's not something we want in the future I think.

If you want to iterate on that Kotlin scripts support soon, I suggest that we do a little brainstorming session on the phone, or IRL when I'll be in Paris on the 3rd of November, and I write an issue or a design doc about the considerations for a first version of the feature. That should help us avoid a long to review and fix PR like what happened for versions catalog support.

jmfayard:
@LouisCAD I commented on your idea to have something independant from Gradle here https://github.com/jmfayard/refreshVersions/issues/586

I still think there is no issue with adding Kotlin Script support in the current codebase. 
We can refactor the code later.

Something I need to clarify is that I do **not** want to implement Kotlin Scripts support in one large pull request here like I did for Gradle Versions Catalogs. 
I learned my lesson that small is beautiful, and having multiple small pulll requests is less work and less stress for everyone that one huge pull request that does everything.
So this particular PR has the narrow goal of **adding the command-line flag and discussing how the feature should look like** but nothing more.

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/RefreshVersionsCleanupTask.kt
jmfayard: actually no, see comment :)

LouisCAD:
I don't think that getting this PR merged or not is an actual blocker to start working/iterating on the kotlin scripts support feature, but since we might want to add further feature flags at some point, which would generate little git conflicts, I guess it's a good idea to merge it now.

PR #622, opened 2022-09-21
[Issue-621] Updated docs for Import -> Link Gradle Project terminology

thederputy:
Hello/bonjour, this is a minor documentation update to the dev set up page.

## What?

Resolves #621

Minor documentation update to account for the change in terminology in newer IntelliJ versions (since at least 2022.1)

## Why?

To help new contributors to the project.

## How?

Took a new screenshot on the latest IntelliJ

## Testing?

I have yet to test the changes, but I'm pushing this early so that @LouisCAD can review for the next release.

PR #623, opened 2022-09-22
Reduced image size of docs/img/*.png by using TinyPNG

thederputy:
Hello/bonjour, based on the image change in #622 I've gone ahead and reduced the image size on all PNGs in the docs/img folder

## What?

Reduced the size of PNGs in `/docs/img/`

## Why?

Saving space is a good thing!

## How?

Uploaded images to [TinyPNG](https://tinypng.com/) and downloaded the resulting "tinified" images.

## Testing?

Tested the size savings for all images. The one `.jpg` did not see any size savings therefore it is not included.


LouisCAD (approved):
Hi, I'm sorry, I don't see the changes…

That's good, thank you for that sizable size reduction!

PR #646, opened 2022-11-24
Update Mockito kotlin package

ken-kentan:
Hello/bonjour,

## What?

I update `Testing.kt`'s package reference from old to new package name

## Why?

The package `com.nhaarman.mockitokotlin2:mockito-kotlin` was moved to `org.mockito.kotlin:mockito-kotlin`

## How?

Update group name from `com.nhaarman.mockitokotlin2` to `org.mockito.kotlin`

## Testing?

I run `gradle check` in `/plugins` folder.


plugins/dependencies/src/main/kotlin/dependencies/Testing.kt
```diff
@@ -255,12 +253,7 @@ object Testing : IsNotADependency {
         val inline = module("mockito-inline")
         val errorProne = module("mockito-errorprone")
         val junitJupiter = module("mockito-junit-jupiter")
-
-        /**
-         * Using Mockito with Kotlin
-         * [More info here](https://github.com/nhaarman/mockito-kotlin)
-         */
```
ken-kentan, line 262: I suppose that we don't need any additional comment since `mockito-kotlin` is part of offficial Mockito library now.

PR #657, opened 2023-01-24
:sparkles: Add Kotlin Gradle Plugin dependency

MaxMichel2:
Hello/bonjour,

## What?

Adds Kotlin Gradle Plugin dependency

## Why?

Quality of life for Android developers (Why have only AGP and not Kotlin?)

## How?

Added module to Kotlin Dependency Group

## Testing?

Ran `check` on both `sample-android` and `sample-groovy`


LouisCAD (approved):
I agree, I actually missed it myself several times.
Thank you!

PR #670, opened 2023-02-09
fix noop comparison

Vampire:
## What?

Fixes a noop comparison.

## Why?

Because it is unlikely that it was intended that way.

## How?

Do the right comparison

## Testing?

I did not as I'm not sure how the effect of the wrong comparison is,
other than that it searches the lowest version for all dependencies
with that group in the version catalog instead of exactly that dependency
coordinates. I verified by looking at the debugger.


LouisCAD (approved):
That was a sneaky one, thank you!

PR #671, opened 2023-02-09
npm libs in version catalogs

Vampire:
## What?

Handle version for NPM libs that are defined in a version catalog.

## Why?

So that you get version update comments for NPM libs in the version catalog.
How to define and use them can be seen at https://youtrack.jetbrains.com/issue/KT-56416/KJS-Gradle-npm-compatibility-with-Gradle-version-catalogs

## How?

Most things were in place already, I just needed to adjust 3 spots to handle NPM dependencies accordingly.

## Testing?

Manually with my setup-wsl project.


plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/extensions/gradle/Dependency.kt
```diff
@@ -23,5 +23,5 @@ internal fun Dependency.npmModuleId(): ModuleId.Npm {
 }
 
 internal fun Dependency.matches(moduleId: ModuleId): Boolean {
-    return moduleId.group == group && moduleId.name == name
```
LouisCAD, line 26: This leads to an allocation on every single `matches` check, that's the reason why I didn't just use the conversion trick.

Could you revert this change?

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/VersionManagementKind.kt
```diff
@@ -89,15 +91,28 @@ private fun Dependency.hasVersionInVersionCatalog(
     versionsCatalogLibraries: Collection<MinimalExternalModuleDependency>,
     versionsCatalogPlugins: Set<PluginDependencyCompat> = emptySet()
 ): Boolean {
-    if (this !is ExternalDependency) return false
+    when {
+        this::class.simpleName == "NpmDependency" -> {
+            return versionsCatalogLibraries.any {
+                val moduleId = npmModuleId()
+                it.module.group == (moduleId.group ?: "<unscoped>") && it.module.name == moduleId.name
```
LouisCAD, line 98: This "`<unscoped>`" thing is a a convention you defined, that isn't standard, right?
Isn't it possible to just have no group (i.e. `null`) instead?

LouisCAD (changes requested):

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/VersionManagementKind.kt
Vampire: Yes, it is a convention I invented and also suggested in the mentioned https://youtrack.jetbrains.com/issue/KT-56416/KJS-Gradle-npm-compatibility-with-Gradle-version-catalogs ticket.
I hope for it to become the official standard once that ticket maybe gets attention.

Just having `semver = { module = "semver", version.ref = "semver" }` does not work, as the version catalog parser complains it is not valid syntax.
Same for `semver = { module = ":semver", version.ref = "semver" }`, invalid syntax.
So you need the colon and you need something before it.
`semver = { module = "null:semver", version.ref = "semver" }` would be valid in the version catalog, but actually `"null"` would be a valid scope too and could also easily confuse users, so I decided to use something that clearly says it is an unscoped package and is very unlikely to be used as actual scope if it is even a valid scope name, hence `"<unscoped>"`.

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/extensions/gradle/Dependency.kt
Vampire: No I can not, otherwise it does not work anymore, that was not just a cosmetic change:
```plain
moduleId.group == group && moduleId.name == name
actions == null && cache == @actions/cache
moduleId == moduleId()
Npm(group=actions, name=cache) == Npm(group=actions, name=cache)
```
I can try to reprogram it so that still the necessary calculation is done but no allocation necessary maybe.

plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/VersionManagementKind.kt
LouisCAD: What I would hope is for Gradle to allow `null` group as they do when not using version catalogs, that'd cut down the noise. On the other hand, unscoped dependencies are not that great of a thing, so maybe making it super obvious might be a good thing… 🤔

PR #693, opened 2023-07-18
fix broken url in PULL_REQUEST_TEMPLATE

uberbinge:
Hello/bonjour,

## What?

fix broken url which is part of the pull request template


LouisCAD (approved):
Thanks!

PR #697, opened 2023-08-21
Add missing '$' before {{  secrets.GITHUB_TOKEN }} in refreshversions-bot.md doc

mr3y-the-programmer:
Hello/bonjour,

## What?

Fix yaml syntax error by adding '$' before "{{  secrets.GITHUB_TOKEN }}" in [RefreshVersionsBot doc](https://splitties.github.io/refreshVersions/refreshversions-bot/#yaml-workflow)

## Why?

because otherwise the execution of refresh versions workflow fails with an error message like this: 
```
The workflow is not valid. .github/workflows/refreshVersions.yml (Line: 33, Col: 25): A mapping was not expected .github/workflows/refreshVersions.yml (Line: 57, Col: 25): A mapping was not expected
```

## How?

Add '$' before "{{  secrets.GITHUB_TOKEN }}"

### DIFF
--- a/docs/refreshversions-bot.md
+++ b/docs/refreshversions-bot.md
@@ -46,7 +46,7 @@ jobs:
         with:
           branch: dependency-update
         env:
-          GITHUB_TOKEN: {{ '{{' }} secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


LouisCAD:
Thank you!

PR #720, opened 2024-05-15
:sparkles: Update Android sample

MaxMichel2:
Hello/bonjour,

## What?

Updates the Android sample to function correctly with the Gradle 8+ requirement

## Why?

Because this should have been done during the migration to Gradle 8+

## How?

Simply updated the Gradle Wrapper and the associated files.

## Testing?

Tried to sync the android sample and if it works, the test is validated.


MaxMichel2:
@LouisCAD or @jmfayard, can either of you review/approve this PR ?

PR #724, opened 2024-08-28
Bugfix/version catalog writing

### LINKED ISSUE #705: Version alignement issue
- [x] I have read the [guidelines for submitting issues](https://splitties.github.io/refreshVersions/contributing/submitting-issues/#bug-reports)
## 🐛 Describe the bug

The alignement of the versions in the version catalog is incorrect

## ⚠️ Current behavior

All available versions are misaligned with the currently defined one by one space, see image below

image

## ✅ Expected behavior

The available versions should be aligned with the currently defined version instead

## 💣 Steps to reproduce

Just implement refreshVersions in a project

## 📱 Tech info
 - [Build Scan URL](https://dev.to/jmfayard/the-one-gradle-trick-that-supersedes-all-the-others-5bpg):

MaxMichel2:
Hello/bonjour,

## What?

Resolves #705

* Changed the symbol used for available versions in the version catalog
* Changed the TOML sections order for the version catalog

## Why?

Fixes the issues with version alignment in the version catalog which was dependent on the font used in the IDE. The symbol used is now font-agnostic and is always the same size.

This PR also fixes the somewhat annoying issue where the sections of the version catalog were re-ordered. The order now follows what most people will expect and also what is described in the Gradle documentation (See [here](https://docs.gradle.org/current/userguide/platforms.html#sub::toml-dependencies-format)).

## How?

* Replaced the `⬆` symbol with the `^` character
* Set the TOML sections order to the following:
    * Root
    * `[versions]`
    * `[libraries]`
    * `[bundles]`
    * `[plugins]`

## Testing?

Published to mavenLocal and tested in a local project using refreshVersions

### DIFF
--- a/plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/TomlSection.kt
+++ b/plugins/core/src/main/kotlin/de/fayard/refreshVersions/core/internal/TomlSection.kt
@@ -6,13 +6,13 @@ internal sealed class TomlSection(open val name: String) {
 
     object Root: TomlSection("root")
     object Versions: TomlSection("versions")
-    object Plugins: TomlSection("plugins")
-    object Bundles: TomlSection("bundles")
     object Libraries: TomlSection("libraries")
+    object Bundles: TomlSection("bundles")
+    object Plugins: TomlSection("plugins")
     data class Custom(override val name: String) : TomlSection(name)
 
     companion object {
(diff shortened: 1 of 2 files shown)


MaxMichel2:
FYI, the check has failed due to some signing configuration that doesn't seem to work (This hasn't been modified so I suspect the issue lies on your side since I have no way of handling this)

PR #741, opened 2026-07-30
Keep version-less dependency declarations untouched in refreshVersionsMigrate

### LINKED ISSUE #740: refreshVersionsMigrate appends the version placeholder to version-less dependency declarations, breaking BoM/platform-based builds
- [x] I have read the [guidelines for submitting issues](https://splitties.github.io/refreshVersions/contributing/submitting-issues/#bug-reports)

## 🐛 Describe the bug

`refreshVersionsMigrate` appends the version placeholder to **version-less** dependency declarations, i.e. declarations whose version is intentionally provided by a platform/BoM (Spring Boot dependency management, `aws.sdk.kotlin:bom`, etc.) or by dependency constraints:

```diff
-    implementation("com.fasterxml.jackson.module:jackson-module-kotlin")
+    implementation("com.fasterxml.jackson.module:jackson-module-kotlin:_")
```

This breaks the build: the migration only adds `versions.properties` entries for dependencies that have a hardcoded version (`addMissingEntriesInVersionsProperties` filters on `hasHardcodedVersion`), so the newly written placeholders have no matching entry and can never resolve:

```
* What went wrong:
Execution failed for task ':buildSrc:compileKotlin'.
> Could not resolve all files for configuration ':buildSrc:compileClasspath'.
   > Could not find com.fasterxml.jackson.module:jackson-module-kotlin:_.
```

Since the rewrite is a purely textual regex replacement, it also corrupts string literals that merely *look like* coordinates but are not dependency declarations, silently disabling them. Real example from our build, where a `resolutionStrategy.eachDependency` matcher stops matching forever:

```diff
 resolutionStrategy.eachDependency {
     when ("${requested.group}:${requested.name}") {
-        "co.elastic.clients:elasticsearch-java" -> useVersion(elasticsearchJavaVersion)
+        "co.elastic.clients:elasticsearch-java:_" -> useVersion(elasticsearchJavaVersion)
```

The cause is the empty alternative in `mavenCoordinateRegex`, which makes it match coordinates without a version, combined with the fallback that unconditionally appends `:_`:

https://github.com/Splitties/refreshVersions/blob/2a8be134281c5a5f4b1cbc9052b1a48a1a56761e/plugins/dependencies/src/main/kotlin/de/fayard/refreshVersions/RefreshVersionsMigrateTask.kt#L253-L255

## ⚠️ Current behavior

Running `./gradlew refreshVersionsMigrate --mode=VersionsPropertiesOnly` on a project that uses a BoM/platform rewrites every version-less coordinate to `coordinate:_` and leaves the build broken (`Could not find <group>:<name>:_.`).

## ✅ Expected behavior

Version-less dependency declarations are left untouched by the migration:

- In Gradle, a version-less declaration only resolves when a platform/BoM or dependency constraint supplies the version, so it is a reliable signal that the version is deliberately managed outside of refreshVersions.
- Appending `:_` to such a declaration can never help: no `versions.properties` entry is created for it, so the placeholder cannot resolve and the build fails.

Replacing version-less coordinates with built-in dependency notations when available (covered by the existing `"Replace with dependency names, if present"` test with `org.apache.logging.log4j:log4j-jul`) is fine and should keep working — only the `:_` fallback should skip them.

## 💣 Steps to reproduce

1. Create a project that declares a dependency without a version, resolved through a BoM, e.g.:

    ```kotlin
    dependencies {
        implementation(platform("aws.sdk.kotlin:bom:_"))
        implementation("aws.sdk.kotlin:s3-jvm")
    }
    ```

2. Run `./gradlew refreshVersionsMigrate --mode=VersionsPropertiesOnly`
3. The declaration becomes `implementation("aws.sdk.kotlin:s3-jvm:_")`, no `versions.properties` entry is added for it, and any subsequent Gradle invocation fails with `Could not find aws.sdk.kotlin:s3-jvm:_.`

## 📱 Tech info

- refreshVersions version: 0.60.5 (also reproduced on current `main`, the relevant code is unchanged)
- Gradle: 8.14.3
- OS: macOS / Linux (not OS-specific)

I'll open a PR with a fix and a test shortly.

takaharu-nakase:
Fixes #740

## Problem

`refreshVersionsMigrate` appends the version placeholder to **version-less** dependency declarations:

```diff
-    implementation("com.fasterxml.jackson.module:jackson-module-kotlin")
+    implementation("com.fasterxml.jackson.module:jackson-module-kotlin:_")
```

A version-less declaration means the version is supplied by a platform/BoM (Spring Boot dependency management, `aws.sdk.kotlin:bom`, …) or by dependency constraints. Migrating it to `:_` always breaks the build, because the migration only adds `versions.properties` entries for dependencies with hardcoded versions, so the new placeholder can never resolve:

```
> Could not find com.fasterxml.jackson.module:jackson-module-kotlin:_.
```

Because the rewrite is textual, it also corrupts string literals that merely look like coordinates, e.g. `resolutionStrategy.eachDependency` matchers (`"co.elastic.clients:elasticsearch-java" -> useVersion(…)` becomes `"co.elastic.clients:elasticsearch-java:_" -> …`, which never matches again).

The cause is the empty alternative in `mavenCoordinateRegex` (`(?:|:_|:$versionChars{3,})`), which makes the fallback replacement match coordinates without a version.

## Fix

Introduce `mavenCoordinateWithVersionRegex` (same regex, without the empty alternative) and use it for the `:_` fallback replacement only. Version-less coordinates are now left untouched by the fallback.

The replacement of version-less coordinates with built-in dependency notations, when a mapping exists, is intentionally kept as-is (behavior covered by the existing `"Replace with dependency names, if present"` test via `org.apache.logging.log4j:log4j-jul`).

## Testing

- New test case `"Keep version-less maven coordinates untouched in build files"` in `MigrationTest`
- `./gradlew :refreshVersions:test --tests "de.fayard.refreshVersions.MigrationTest"` passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)


takaharu-nakase:
### CI failure analysis: pre-existing on `main`, not caused by this change

The `check-all` job fails at the **"Check for memory leaks"** step, which stress-runs the `sample-groovy` build 10 times in a daemon capped at `-Xmx150M`. The OOM happens on **run number 1**, during Kotlin DSL script compilation of `sample-groovy/buildSrc/build.gradle.kts` — before any repeated-build accumulation could show up:

```
Run number 1
e: Java heap space (build.gradle.kts): java.lang.OutOfMemoryError: Java heap space
```

To rule this PR out, I dispatched the same workflow on **unmodified `main`** in my fork, and it fails identically at the same step, also on run number 1:

- This PR: https://github.com/Splitties/refreshVersions/actions/runs/30515345150
- Unmodified `main` (control run): https://github.com/takaharu-nakase/refreshVersions/actions/runs/30515593606

The last green runs of this workflow were on 2026-07-01 (PR #739's re-run) and 2025-08-16 (`main`), so it looks like the current GitHub-hosted runner environment now needs more than 150 MiB for Gradle 9 script compilation, which makes the step fail for any PR regardless of its content.

I'm verifying a heap bump for the stress test on my fork and will send a companion PR if it goes green.

takaharu-nakase:
I've cherry-picked the CI fix from #742 onto this branch so CI can actually run for this PR — happy to drop that commit once #742 lands. Verification of the workflow fix on unmodified `main`: https://github.com/takaharu-nakase/refreshVersions/actions/runs/30515882804

